PR #36, opened 2022-12-16
Add consts/text for Gen 9 abilities

Drayano60:
Some of these abilities have too many effects to squeeze into two lines <__>

(Assuming it's fine to add these, if you want to keep Gen 9 stuff out for now dw)


data/text/720.txt
```diff
@@ -265,4 +265,35 @@ Dragon’s Maw
 Chilling Neigh
 Grim Neigh
 As One
-As One
\ No newline at end of file
+As One
```
Drayano60: I'm not sure what's with the diff here tbh

Drayano60:
After some experimentation this is probably a bit premature; there's still only one byte for each ability in the personal data I think so abilities with an ID of 256 or over don't work

BluRosie:
yeah, need to edit the GetBoxMonData function to grab the ability from either a different field or add one of the bits from the same byte we put the hidden ability thing

BluRosie:
that would probably still have bugs with fields expecting a u8 so that'd be a second i believe

PR #38, opened 2022-12-23
Renumber species to add cross-gen evolutions after Gen 4

Drayano60:
I'm only planning to use Gen 4 + cross-gen evos so this keeps them closer together in the Pokédex.


Drayano60:
Sorry it messed up when I was selecting the base branch lol, ignore this >__>

PR #53, opened 2023-03-22
ported transparent textboxes and made it an optional setting with default mode set to off

turtleisaac:
ported from https://github.com/BluRosie/hg-transparent-textbox


turtleisaac:
Main thing that needs to be addressed is the question of whether or not a config option is needed in the makefile to prevent the contents of the narc files from being copied in. Right now when the option in include/config.h is turned off, the narc subfiles still get copied in. Tests show at the surface level that nothing goes wrong but I’m not trusting that without verification

PR #61, opened 2023-04-04
First run at an import script for PokEditor sheets

lhearachel:
This is just the easy stuff: Personal, Level-Up Learnsets, and Evolutions.

Priority stuff that still needs to be done:
* Trainers
* Moves

I tested these functions out on a set of sheets generated from a clean copy of vanilla HG and inspected for diffs from there. This script *does* break the Fairy implementation config flag, but if a user is using this script, they probably already know if they want that or not.


BluRosie:
let me know if you need help with the trainers and moves 😁

PR #67, opened 2023-04-11
Full Metal Body cannot be ignored

lhearachel:
Also includes a small fix in the Defiant check to pass `bw` instead of `sp` to a BattleWork function.


src/battle/battle_script_commands.c
```diff
@@ -828,7 +828,7 @@ BOOL btl_scr_cmd_33_statbuffchange(void *bw, struct BattleStruct *sp)
                 }
                 else if ((MoldBreakerAbilityCheck(sp, sp->attack_client, sp->state_client, ABILITY_CLEAR_BODY) == TRUE)
                       || (MoldBreakerAbilityCheck(sp, sp->attack_client, sp->state_client, ABILITY_WHITE_SMOKE) == TRUE)
-                      || (MoldBreakerAbilityCheck(sp, sp->attack_client, sp->state_client, ABILITY_FULL_METAL_BODY) == TRUE)
+                      || (sp->battlemon[sp->state_client].ability == ABILITY_FULL_METAL_BODY)   // Full Metal Body cannot be ignored
```
BluRosie, line 831: do a `GetBattlerAbility(sp, sp->state_client)` here instead

BluRosie (changes requested):

BluRosie (approved):

PR #72, opened 2023-04-24
Fix wild level-up moves of alternate forms

chikorita47:
I have not tested this in HG, but in Pt there is an issue with wild alternate forms, which is that they get moves from both the base form and the alternate form learnset.

The reason is that, at least in Pt, in most cases (unless, I think, Cute Charm is present) CreateBoxMonData is called before AddWildPartyPokemon. So CreateMonBoxData initializes the moveset with original form moves, form is set in AddWildPartyPokemon, which then overrides the moves with the alternate form learnset. But if the alternate form doesn't have 4 moves to learn, the base form moves are still there. For example, wild lv3 Alolan Vulpix has both Ember and Powder Snow.

This was made worse by https://github.com/BluRosie/hg-engine/commit/b9389299ace40ab204677051512a7855223930df, which now sets the form to 0 for all mons besides those listed in the function. This means the form set here is immediately overridden: https://github.com/BluRosie/hg-engine/blob/b9389299ace40ab204677051512a7855223930df/asm/other_hook.s#L222-L231 and so the base form moveset is initialized AFTER the alternate form (meaning it will always be used --> A-Vulpix always has normal Vulpix's moves).

These changes fix the issue in Pt --> wild lv3 Alolan Vulpix has only Powder Snow and Tail Whip, as expected. Again, have not tested in HG, I think these changes should be safe though?

### DIFF
--- a/src/pokemon.c
+++ b/src/pokemon.c
@@ -3172,6 +3172,7 @@ void __attribute__((long_call)) UpdatePassiveForms(struct PartyPokemon *pp)
 {
     u32 species = GetMonData(pp, ID_PARA_monsno, NULL);
     u32 form = 0;
+    BOOL shouldUpdate = TRUE;
     
     switch (species)
     {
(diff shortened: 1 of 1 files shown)


BluRosie:
looks good to me!  thank you so much!

PR #74, opened 2023-05-01
added the defender item effect check on hit function

turtleisaac:


src/battle/battle_item.c
```diff
@@ -179,3 +184,303 @@ u32 ServerWazaHitAfterCheckAct(void *bw, struct BattleStruct *sp)
 
     return (ret == 1);
 }
+
+
+//thanks to Lhea for this function - TODO ask for their SUB_SEQ scripts for the gen5+ item effects since rn those can't be used
+//this function is for a defender's held item effect triggering after being hit
+//go to ServerWazaHitAfterCheckAct for implementing an attacker's held item effect triggering after using a move
```
lhearachel, line 191: For reference, I need to actually implement these SUB_SEQ scripts myself.

turtleisaac: lmao

PR #90, opened 2023-06-14
battery, power spot, steely spirit, steelworker & rocky payload

mozzydippers:


BluRosie:
looks good now, thank you for the contribution!  i will merge it in.

i will note that currently abilities above index 255 (a little before the end of gen 8) do not function.  it will be a massive overhaul to get them to function, so i'm kinda procrastinating on working on them.  but the code contributions are still necessary, so thank you very much 😁

PR #92, opened 2023-06-18
Add effects for Bulletproof, Competitive, Gale Wings, Berserk, Fluffy, Merciless

Drayano60:
Just porting over some of the abilities I've done in my own fork that I think have a fairly solid implementation. I think these should all work as expected, with the slight exception that Berserk can activate in the middle of a multi-hit move instead of at the end.

I have a few more Gen 6-9 abilities in my own fork (Dazzling/Queenly Majesty/Cotton Down/Quick Draw/Armor Tail/Supreme Overlord) but their implementation is a bit scruffy or not fully accurate so figured best not include those!

There's a separate commit for each ability if you want to see the specific code isolated.


BluRosie:
will review this in a bit!

BluRosie:
looks good to me!

PR #94, opened 2023-06-19
Giving names to (nearly) all the battle script vars

PandaPanda-Panda:
Name changes / pointer locations obtained from decomp of function at 0x02245528, overlay 12.
 
Still unsure what Var_70 points to, near the end of the BattleSystem struct (offset 0x247C).
In the field condition struct, changed to English move names


BluRosie:
this is huge 😁  will review in a bit.

PR #97, opened 2023-06-21
Beat Up(date) and Friend ball fix

PandaPanda-Panda:
Updated beat up's damage calc.
Sneasels learnset level for beat up doesn't correspond to any mainline game so adjusted it to Gen IV's standard.
It seems the friend ball was missed in the handling for ball shakes/crit captures, so that's been added to the end.


include/battle.h
```diff
@@ -1250,7 +1251,8 @@ BOOL __attribute__((long_call)) HeldItemHealStatusCheck(void *bw, struct BattleS
 int __attribute__((long_call)) TypeCheckCalc(struct BattleStruct *sp, u32 attack_client, u32 typeModifier, int damage, int base_power, int *flag);
 BOOL __attribute__((long_call)) ShouldDelayTurnEffectivenessChecking(struct BattleStruct *sp, u32 move_no);
 BOOL __attribute__((long_call)) ShouldUseNormalTypeEffCalc(struct BattleStruct *sp, int attack_client, int defence_client, int pos);
-
+int __attribute__((long_call)) GetPartySize(void *bw, int client_no);
+int __attribute__((long_call)) GetPartyMon(void *bw, int client_no, int mon_index);
```
BluRosie, line 1255: make these like `Battle_GetClientPartySize` and `Battle_GetClientPartyMon` because there are other functions for generic `struct Party`'s that would better use these names

src/battle/battle_script_commands.c
```diff
@@ -1186,6 +1186,76 @@ BOOL btl_scr_cmd_54_ohko_move_handle(void *bw, struct BattleStruct *sp)
...
+        }
+    }   
+
+    mon = GetPartyMon(bw, sp->attack_client, sp->beat_up_count);
+    species = GetMonData(mon, ID_PARA_monsno, 0);
+    form = GetMonData(mon, ID_PARA_form_no, 0);
+
+    newBaseDamage = PokeFormNoPersonalParaGet(species, form, PERSONAL_BASE_ATTACK);
+    newBaseDamage /= 10;
+    sp->damage_power = sp->aiWorkTable.old_moveTbl[sp->current_move_index].power;
```
BluRosie, line 1225: this should be `sp->moveTbl[sp->current_move_index].power` due to expanded move table being moved out of the aiWorkTable structure

src/battle/battle_script_commands.c
```diff
@@ -1646,6 +1716,9 @@ u32 CalculateBallShakes(void *bw, struct BattleStruct *sp)
 
 #endif
     }
+    if(sp->item_work = ITEM_FRIEND_BALL && i > 4){
+        SetMonData(GetPartyMon(bw,sp->defence_client,0),9,200);
+    }
```
BluRosie, line 1721: needs 
```c
if (sp->item_work == ITEM_FRIEND_BALL && i >= 4)
{
    SetMonData(GetPartyMon(bw,sp->defence_client,0), ID_PARA_friend, 200);
}
```
also needs to be handled for captures which are certain to pass (the if... else right above this)

armips/data/moves.s
```diff
@@ -4026,7 +4026,7 @@ movedata MOVE_WHIRLPOOL, "Whirlpool"
 movedata MOVE_BEAT_UP, "Beat Up"
     battleeffect 154
     pss SPLIT_PHYSICAL
-    basepower DEBUG_NEEDS_TESTING ? 0 : 10 // will beat up still work with base power 0?
+    basepower 5
```
BluRosie, line 4029: does this really have base power 5?

src/battle/battle_script_commands.c
```diff
@@ -1186,6 +1186,76 @@ BOOL btl_scr_cmd_54_ohko_move_handle(void *bw, struct BattleStruct *sp)
...
+        sp->multi_hit_count_temp = 2;
+        sp->loop_hit_check = 0xFD;
+        sp->beat_up_count = 0;
+        mon = GetPartyMon(bw, sp->attack_client, sp->beat_up_count); 
+
+        while(!(sp->beat_up_count != sp->sel_mons_no[sp->attack_client] ||
+                (GetMonData(mon, ID_PARA_hp, 0) && 
+                GetMonData(mon, ID_PARA_monsno_egg, 0) && 
+                GetMonData(mon, ID_PARA_monsno_egg, 0) != 494 && 
+                GetMonData(mon, ID_PARA_condition, 0) == 0)))
```
BluRosie, line 1210: that second side of the `||` is checking for if hp is not zero, if the species is not zero, if the species is not an egg, and if the nonvolatile status conditions are nonzero

BluRosie (changes requested):
got a few things to address here.  let me know if you have any questions!

armips/data/moves.s
PandaPanda-Panda: image
Not 5 exactly, for each mon in the party it's their base attack / 10 with 5 added to it. Like fury cutter, the move's damage calc adjusts the base damage before the main damage calculation

BluRosie: what does it show in the summary screen?  this would show 5 i believe--unless i'm wrong and it won't display sub 5

src/battle/battle_script_commands.c
PandaPanda-Panda: Got the logic of this check confused, corrected to
```
while(sp->beat_up_count != sp->sel_mons_no[sp->attack_client] &&
                (GetMonData(mon, ID_PARA_hp, 0) == 0 || 
                GetMonData(mon, ID_PARA_monsno_egg, 0) == 0 || 
                GetMonData(mon, ID_PARA_monsno_egg, 0) == 494 || 
                GetMonData(mon, ID_PARA_condition, 0) != 0))
```

src/battle/battle_script_commands.c
PandaPanda-Panda: Added a copy of this section without `&& i >= 4` to the end of the critical captures shake count calculation to catch single shake captures.

include/battle.h
PandaPanda-Panda: Done

src/battle/battle_script_commands.c
PandaPanda-Panda: Done

armips/data/moves.s
PandaPanda-Panda: You're correct, I didn't think about the summary screen. A quick test shows it works with 0 set, adding 5 in the damage calc.

src/battle/battle_script_commands.c
```diff
@@ -1636,6 +1707,9 @@ u32 CalculateBallShakes(void *bw, struct BattleStruct *sp)
         if (criticalCapture) // succeeded the one chance it had
             i = i | 0x80; // change the flow of the ball callback to make sure that critical captures only shake once then succeed.  if it shakes, it succeeds, though
 
+        if(sp->item_work = ITEM_FRIEND_BALL) // To catch a 1 shake capture
+            SetMonData(Battle_GetClientPartyMon(bw,sp->defence_client,0),ID_PARA_friend,200);
+
```
BluRosie, line 1712: equivalence checks in C are `==`

additionally, the 3rd arg for `SetMonData` needs to be a pointer to the data.  not 100% on why GF handles it that way, but hey:
```c
if (sp->item_work == ITEM_FRIEND_BALL)
{
    u32 friendship = 200;
    SetMonData(Battle_GetClientPartyMon(bw,sp->defence_client,0), ID_PARA_friend, &friendship);
}
```

src/battle/battle_script_commands.c
```diff
@@ -1646,6 +1720,10 @@ u32 CalculateBallShakes(void *bw, struct BattleStruct *sp)
 
 #endif
     }
+
+    if(sp->item_work = ITEM_FRIEND_BALL && i >= 4) 
+        SetMonData(Battle_GetClientPartyMon(bw,sp->defence_client,0),ID_PARA_friend,200);
+
```
BluRosie, line 1726: same as the other friend ball check

src/battle/battle_script_commands.c
BluRosie: also i don't think this is the right place to be inserting this--should only be inserting it for success conditions.  critical captures are not 100% successful

perhaps insert it above the check immediately above with the `criticalCapture` and check if `i == caughtMons`:


https://github.com/BluRosie/hg-engine/blob/b7aa067f2a7e35bacc596b8db3a00696c6105cb9/src/battle/battle_script_commands.c#L1710-L1711

```c
if (sp->item_work == ITEM_FRIEND_BALL && i == caughtMons)
{
    u32 friendship = 200;
    SetMonData(Battle_GetClientPartyMon(bw,sp->defence_client,0), ID_PARA_friend, &friendship);
}
```

also still need to insert it here https://github.com/BluRosie/hg-engine/blob/b7aa067f2a7e35bacc596b8db3a00696c6105cb9/src/battle/battle_script_commands.c#L1660-L1661

src/battle/battle_script_commands.c
BluRosie: can delete this if the other friend ball check is implemented

BluRosie (changes requested):

src/battle/battle_script_commands.c
PandaPanda-Panda: This friend ball check catches both capture rate > 255 and the regular random checks if crit capture isn't defined. I think it's better to keep this here than move it up, for completeness if anyone disables crit captures.

BluRosie: my bad, you're totally right 😁

src/battle/battle_script_commands.c
PandaPanda-Panda: Not sure why but I was processing the break statements in the for loop about as returns, thinking by the time it got the if statement the throw was successful. I'll check i against caughtMons.

BluRosie (approved):
looks good to me, will move the second friend ball check into the `else` so it doesn't activate on unsuccessful critical captures (which would have i = 0x80 which is more than 4)

thank you for the contribution again 😁

PandaPanda-Panda:
Wouldn't it be better to check against i masked with 0xF without moving it so it still catches when capture rate > 255 as that would be skipped in the else?

BluRosie:
that's what i ended up doing in 60d9f06163b0a8b105687c2cf8cff04778d3cea5

PandaPanda-Panda:
I should have checked, thanks for reviewing and correcting my oversights Blu

PR #98, opened 2023-06-21
liquid voice, neuroforce, prism armor & shadow shield

mozzydippers:


src/battle/battle_pokemon.c
```diff
@@ -970,6 +970,10 @@ u32 GetAdjustedMoveTypeBasics(struct BattleStruct *sp, u32 move, u32 ability, u3
     {
         typeLocal = TYPE_NORMAL;
     }
+    else if ((ability == ABILITY_LIQUID_VOICE) && (SoundproofMoveList[i] == sp->current_move_index)))
+    {
+        typeLocal = TYPE_WATER;
+    }
```
BluRosie, line 976: need a for loop to check through `SoundProofMoveList`--`i` as a variable isn't initialized here either so it won't compile.  even if it were it would just check 1 index instead of the full `SoundProofMoveList`

[see how strong jaw is handled](https://github.com/BluRosie/hg-engine/blob/main/src/battle/battle_calc_damage.c#L735-L742)

BluRosie (changes requested):

src/battle/battle_pokemon.c
mozzydippers: done !

BluRosie:
thank you again for your contribution!  😁

PR #147, opened 2023-11-09
Move Animations

MichaelFacci:
Scripted animations for Scald, Electroweb, Inferno, Night Daze, Snarl, Frost Breath, Dual Chop, Bulldoze, Volt Switch, and Icicle Crash


BluRosie:
i will review this shortly, thank you!

Issue #135, opened 2023-10-03
move selection sometimes crashes in multi battles for player's left


BluRosie:
so i have this currently happening in two different scenarios:

diggersby on left, anything on right (specifically landorus in tests comes to mind)
diggersby can not select a move else it crashes--does this have to do with therian forme being a thing?

same thing with hisuian zorua in illusion as darmanitan on left, but on the right having like.  sandile also crashes the game on move selection.  wonder if this does have something to do with forms being present

BluRosie:
so this is technically fixed...  but the palette for primal reversion is being overwritten.  may just have to move the tag elsewhere

Issue #121, opened 2023-08-21
weather icons don't like to keep their palettes in double battles


BluRosie:
![image](https://github.com/BluRosie/hg-engine/assets/16446370/8a1a5312-51bb-4517-aa30-380293452bc7)

SergooPoke:
issue encountered in wild double battle -- multi battles untested.

NathanWichers:
end of turn updating also broke

BluRosie:
end of turn updating also appears to have broken

NathanWichers:
ServerFieldConditionCheck

PR #155, opened 2023-11-18
Anger Point | Multi Megas | Dex entries utility script

### LINKED ISSUE #153: Anger Point
- [ ] anger point no longer activates when substitutes are hit

see [subseq 16](https://github.com/BluRosie/hg-engine/blob/main/armips/move/battle_sub_seq/016.s#L17)

### LINKED ISSUE #154: multi battles do not account for megas on individual teams

Mixone-FinallyHere:
Fixes #153 

Fixes #154 

Adds a script that generates a `monData.s` in `scripts/fixed` that checks all defined mon dex entries and in those where newline character is beyond 40 characters corrects to put it at a correct character length, it does not howver use "intelligence" so it simply makes the entry not break, they probably still need editing but at least it allwos to write dex entries without new lines and then auto generate them in a correct format.


BluRosie:
image
here's a fun interaction that fucks with queuing up megas, i'll work on this right quick

BluRosie:
also fixing the last comment in #135 and fixing #121

PR #212, opened 2024-03-04
Modernise battle messages

Aeliko:
Change messages to use "opposing" instead of "foe's"
Modernise other messages to latest standards
image

## TODO:
- [x] Mega Evolution messages
- [x] Distinguish X and Y Mega Evolutions in messages (Edit: GF did not do that so we don't need to)
- [x] Add trainer name in Mega Evolution messages (SM didn't have it so our work is technically done)
- [x] Frisk messages
- [ ] Future proof messages by considering 12-character species/nicknames


BluRosie:
i will probably also use this opportunity to add a difference between moves as they appear in most contexts with the shortened names and moves as they appear when they are used

BluRosie:
going to merge this and then do the battle move messages on the move-dev branch

PR #218, opened 2024-03-17
Implement Field Use of Mirror Herb and Pokémon Nursery Mechanics

Aeliko:
- [x] Core functionality of Mirror Herb. In Gen 9, if a Pokémon holds a Mirror Herb, it will receive Egg Moves from the other regardless of species.
- [x] Deposit order does not matter.
- [x] Inherit more than 1 Egg Move.
- [x] Starting in Gen 8, if two Pokémon of the same species are together in the Pokémon Nursery, one knows an Egg Move, and the other has an empty slot, the other Pokémon will receive the Egg Move in the empty slot.
- [x] Egg Moves are transferred in the same order they appear on the move list ([Egg Move](https://bulbapedia.bulbagarden.net/wiki/Egg_Move)). ~Moves are transferred in descending order on the move list ([Pokémon Nursery](https://bulbapedia.bulbagarden.net/wiki/Pok%C3%A9mon_Nursery#Galar)).~

### DIFF
--- a/hooks
+++ b/hooks
@@ -433,3 +433,5 @@ arm9 MonTryLearnMoveOnLevelUp 02071534 3
 
 # edit the playcry command to potentially take form into account
 0007 AnimCmd_PlayCryEdit_hook 0221F854 0
+
+arm9 ScrCmd_DaycareSanitizeMon 0204CFB8 1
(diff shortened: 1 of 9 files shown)


BluRosie:
will close #231

PR #221, opened 2024-03-27
Modernise Trick Room

Aeliko:
From [Generation V](https://bulbapedia.bulbagarden.net/wiki/Generation_V) onwards, due to the way Trick Room affects Speed calculation, Pokémon with 1809 or more Speed will still act before Pokémon slower than 1809, even in Trick Room. Specifically, Pokémon with a Speed stat closest to 1809 without going under go first. This is because, **under Trick Room, every Pokemon's Speed is set to (10,000 - (their actual Speed)) modulo 8,192**.


BluRosie:
i am interested to see what the >10,000 case is

PR #223, opened 2024-03-29
Implement Dynamic Speed

Aeliko:
In [Generation VII](https://bulbapedia.bulbagarden.net/wiki/Generation_VII), A Pokémon's Speed after [Mega Evolution](https://bulbapedia.bulbagarden.net/wiki/Mega_Evolution) is used to determine the turn order, not its Speed before ([Alterations from Generation VI](https://bulbapedia.bulbagarden.net/wiki/Generation_VII#Alterations_from_Generation_VI)).

In Generation 8, Speed is recalculated mid-turn before each individual Pokémon moves ([Pokémon Sword and Shield – New Competitive Features and Mechanic Changes](https://www.trainertower.com/pokemon-sword-and-shield-new-competitive-features-and-mechanic-changes/)).

### DIFF
--- a/src/battle/other_battle_calculators.c
+++ b/src/battle/other_battle_calculators.c
@@ -907,6 +907,85 @@ u8 CalcSpeed(void *bw, struct BattleStruct *sp, int client1, int client2, int fl
     return ret;
 }
 
+/**
+ *  @brief Sorts clients' execution order factoring in who has already performed their action
+ *  @param bw battle work structure; void * because we haven't defined the battle work structure. Apparently we have but we don't use it here so
+ *  @param sp global battle structure
+ */
+void DynamicSortClientExecutionOrder(void *bw, struct BattleStruct *sp) {
+    int maxBattlers;
+    int i, j;
+    int temp1, temp2;
+    int currentAttackerId = 0;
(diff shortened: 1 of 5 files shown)


src/battle/other_battle_calculators.c
```diff
@@ -907,6 +907,85 @@ u8 CalcSpeed(void *bw, struct BattleStruct *sp, int client1, int client2, int fl
...
+void DynamicSortClientExecutionOrder(void *bw, struct BattleStruct *sp) {
+    int maxBattlers;
+    int i, j;
+    int temp1, temp2;
+    int currentAttackerId = 0;
+
+    maxBattlers = BattleWorkClientSetMaxGet(bw);
+
+    for (i = 0; i < maxBattlers; i++) {
+        if (sp->attack_client == sp->client_agi_work[i]) {
```
BluRosie, line 924: may want to swap `turn_order` as well?  and may also want to add a parameter to the function that specifies the start client potentially (instead of always relying on `attack_client`)?

BluRosie (changes requested):

BluRosie (approved):

Issue #220, opened 2024-03-22
Update Friendship evolution threshold 


Aeliko:
[Friendship evolution](https://m.bulbapedia.bulbagarden.net/wiki/Friendship_evolution)

Generation VIII onwards

From Generation VIII onwards, a Pokémon requires at least 160 friendship in order to evolve.

Relevant code section: [https://github.com/BluRosie/hg-engine/blob/main/src%2Findividual%2FGetMonEvolutionInternal.c#L107](https://github.com/BluRosie/hg-engine/blob/main/src%2Findividual%2FGetMonEvolutionInternal.c#L107)

PR #224, opened 2024-04-01
Update GetMonEvolutionInternal.c

Charliexox:
Changed friendship evolution value from 220 to 160 as part of a modernisation effort.


Mixone-FinallyHere:
Addresses #220

PR #225, opened 2024-04-03
Fix dynamic speed off by 1 error, implement After You and Quash

Aeliko:
- [x] After You
- [x] Quash


src/battle/battle_script_commands.c
```diff
@@ -2891,7 +2894,7 @@ BOOL btl_scr_cmd_F4_isparentalbondactive(void *bw UNUSED, struct BattleStruct *s
 }
 
 /**
- *  @brief script command to jump somewhere if parental bond is currently active beyond mummy
+ *  @brief script command to permeamently change the battle background
```
BluRosie, line 2897: permeamently

BluRosie (approved):

Issue #199, opened 2024-02-05
SwitchInAbilityCheck needs to be rewritten to better reflect later gens


BluRosie:
neutralizing gas
unnerve
all other abilities (including trace)
items?

Aeliko:
- [https://bulbapedia.bulbagarden.net/wiki/User:FIQ/Turn_sequence](https://bulbapedia.bulbagarden.net/wiki/User:FIQ/Turn_sequence)
- [https://bulbapedia.bulbagarden.net/wiki/User:SnorlaxMonster/End-turn_resolution_order](https://bulbapedia.bulbagarden.net/wiki/User:SnorlaxMonster/End-turn_resolution_order)

Aeliko:
End-turn Resolution Order vanilla locations:
| Modern Order | Event                                                                 | HG Location                                      |
|--------------|-----------------------------------------------------------------------|--------------------------------------------------|
| 1            | Weather subsiding                                                     | ServerFieldConditionCheck (exposed)              |
| 2            | Hail damage/sandstorm damage/Rain Dish/Dry Skin/Solar Power/Ice Body  | ServerFieldConditionCheck (exposed)              |
| 3            | Self-curing a status condition due to high Affection                  | Modernize affection                              |
| 4            | Future Sight/Doom Desire                                              | BattleControllerPlayer_UpdateFieldConditionExtra |
| 5            | Wish                                                                  | ServerFieldConditionCheck (exposed)              |
| 6            | Block A                                                               |                                                  |
| 7            | Aqua Ring                                                             | BattleControllerPlayer_UpdateMonCondition        |
| 8            | Ingrain                                                               | BattleControllerPlayer_UpdateMonCondition        |
| 9            | Leech Seed                                                            | BattleControllerPlayer_UpdateMonCondition        |
| 10           | Poison/Poison Heal                                                    | BattleControllerPlayer_UpdateMonCondition        |
| 11           | Burn                                                                  | BattleControllerPlayer_UpdateMonCondition        |
| 12           | Nightmare                                                             | BattleControllerPlayer_UpdateMonCondition        |
| 13           | Curse                                                                 | BattleControllerPlayer_UpdateMonCondition        |
| 14           | Bind/Clamp/Fire Spin/Infestation/Magma Storm/Sand Tomb/Whirlpool/Wrap | BattleControllerPlayer_UpdateMonCondition        |
| 15           | Taunt fading                                                          | BattleControllerPlayer_UpdateMonCondition        |
[...]

PR #238, opened 2024-04-24
Fix #233 and address #199 (hopefully)

### LINKED ISSUE #233: Multi-target moves that faint one battler stop there and don't continue to other battlers
see https://discord.com/channels/446824489045721090/1201213967389966378/1228494110172778536

Aeliko:
Addresses and closes #233

The bug is due to how the state machine handles spread moves from what I understand.

Consider the following Double Battle
```
    C    D
   A    B
```
Where $Speed_{A} > Speed_{B} > Speed_{C} > Speed_{D}$
And $Speed_{C} - 1 = Speed_{D}$

Consider `turn_order` to be like a conveyor belt. On the conveyor belt there are 4 items `[A, B, C, D]`, which have been sorted so that the fastest client appears in front.

Now client A uses String Shot, and hits C (the third element) first before hitting D normally, but `turn_order` is sorted after hitting C. Since C at that instant is slower than D, `turn_order` becomes `[A, B, D, C]`. The conveyor belt moves and we arrive at the forth element, and wow it is C again! So we attack C again.

Next consider client A using Earthquake. `turn_order` is `[A, B, C, D]`. First B is hit, then C (the third element) is hit. Assume C faints. `turn_order` is immediately sorted, and since C fainted it is the slowest, making `turn_order` `[A, B, D, C]`. The conveyor belt moves and we arrive at the forth element, and wow it is C again! So we attack C again. But C has already fainted, so we skip C. Thus the illusion of the processing has stopped is observed in #233.

This PR proposes the fix to be only sorting `turn_order` before a client attacks, and during end-turn resolution order.

P.S. Ligma balls why is it so scuffed lmao

Addresses entry effect order part of #199


BluRosie:
i will go through and test this soon

PR #243, opened 2024-05-09
Updates to a few mechanics

Drayano60:
This PR does the following:
- Hidden Power is always 60 BP (G6+)
- Mud Sport and Water Sport cause the damage to be 1/3 (G5+)
- Ice-types are immune to Sheer Cold (G7+)
- Moody no longer raises accuracy or evasion (G8+)
- Transistor is a 30% boost (G9+)
- Protean works once per switch-in (G9+)

I've had similar code for these on my own branch and it worked fine there, but I've admittedly not tested it thoroughly here

I've done separate commits for each so should be easy to see what's changed for each bit


src/battle/battle_start.c
```diff
@@ -417,10 +417,12 @@ void ServerWazaBefore(void *bw, struct BattleStruct *sp)
             if (sp->battlemon[sp->attack_client].ability == ABILITY_PROTEAN
                 && (sp->battlemon[sp->attack_client].type1 != sp->moveTbl[sp->current_move_index].type  // if either type is not the move's type
                     || sp->battlemon[sp->attack_client].type2 != sp->moveTbl[sp->current_move_index].type)
+                && (sp->battlemon[sp->attack_client].protean_flag == 0) // Protean should activate only once per switch-in
                 && sp->moveTbl[sp->current_move_index].power != 0) // the move has to have power in order for it to change the type
             {
                 sp->battlemon[sp->attack_client].type1 = sp->moveTbl[sp->current_move_index].type;
                 sp->battlemon[sp->attack_client].type2 = sp->moveTbl[sp->current_move_index].type;
+                sp->battlemon[sp->attack_client].protean_flag = 1;
```
kumakuma1, line 425: How about #ifef here to enable Gen9 change

BluRosie (approved):

PR #246, opened 2024-05-15
Implement and Fix Abilites and Moves

Aeliko:
# Abilities

- [x] Desolate Land
- [x] Primordial Sea
- [x] Delta Stream

# Fixed Moves

- [x] Psyblade

# Fixed Abilities

- [x] Protean

# Fixed data

- [x] Mega Rayquaza

# Others

- [x] Make weather-inducing moves/abilities fail


src/battle/other_battle_calculators.c
```diff
@@ -2605,3 +2615,163 @@ void BattleControllerPlayer_UpdateMonCondition(void *bw, struct BattleStruct *sp
...
+    // sprintf(buf, "In BattleController_TryMove\n");
+    // debugsyscall(buf);
+
+    switch (ctx->woc_seq_no) {
+    case TRY_MOVE_STATE_CHECK_VALID_TARGET:
+        ctx->woc_seq_no++;
+        if (ov12_0224B398(bsys, ctx) == TRUE) {
+            return;
+        }
+        //fallthrough
```
BluRosie, line 2734: if you put `FALLTHROUGH` here (macro for `__attribute__((fallthrough))`) it will suppress the warnings

src/battle/other_battle_calculators.c
```diff
@@ -2605,3 +2615,163 @@ void BattleControllerPlayer_UpdateMonCondition(void *bw, struct BattleStruct *sp
...
+ *  @brief checks if the given move should be weakened or not (only prints message)
+ *  @param bw battle work structure
+ *  @param sp global battle structure
+ *  @return TRUE/FALSE
+ */
+BOOL CheckStrongWindsWeaken(struct BattleSystem *bw, struct BattleStruct *sp) {
+    u8 defender_type_1 = BattlePokemonParamGet(sp, sp->defence_client, BATTLE_MON_DATA_TYPE1, NULL);
+    u8 defender_type_2 = BattlePokemonParamGet(sp, sp->defence_client, BATTLE_MON_DATA_TYPE2, NULL);
+    u32 move_type = GetAdjustedMoveType(sp, sp->attack_client, sp->current_move_index);
+    u8 i = 0;
```
BluRosie, line 2681: nitpicky, but (and this is a general review)
all that `u{8|16}` variables tend to do is generate extra lsl/lsr instruction pairs to ensure that the variable is 8/16 bits instead of 32, particularly in loops this adds two instructions of unnecessary overhead

BluRosie:
"Make weather-inducing moves/abilities fail" i.e. under the primal weathers the original weathers can't be set?

Aeliko:
> "Make weather-inducing moves/abilities fail" i.e. under the primal weathers the original weathers can't be set?

Yes

src/battle/other_battle_calculators.c
BluRosie: huh, interesting that the comment does the same thing haha

PR #280, opened 2024-06-25
Add swarm table

TixoRebel:
Add a user-modifiable swarm table


BluRosie:
do you know how this ties in to the pokegear?  i.e. how do i add to this table such that a swarm can happen?

TixoRebel:
Yup, once you add a swarm to the swarm table and encounter table the pokegear will automatically pick it up. I tested it by both modifying swarms and adding new ones and was able to see the changes.

PR #281, opened 2024-06-27
Add repel reuse prompt

TixoRebel:
Still a work in progress pending feedback


BluRosie:
super sorry for getting to this so late.  would you mind throwing the repel that was used into the save so that it persists between reboots?  at the end of the `SAVE_MISC_DATA` would be perfect, could then be accessed by `Sav2_Misc_get(SaveBlock2_get()).repelType` or something.  may also be nice to have logic that queues up repels in order of max/super/normal instead of only queuing up the last one.
let me know if you just want to wash your hands of this and i may be able to handle it then

BluRosie (changes requested):
super sorry for getting to this so late. would you mind throwing the repel that was used into the save so that it persists between reboots? at the end of the [`SAVE_MISC_DATA`](https://github.com/BluRosie/hg-engine/blob/main/include/save.h#L116) would be perfect, could then be accessed by `Sav2_Misc_get(SaveBlock2_get())->repelType` or something. may also be nice to have logic that queues up repels in order of max/super/normal instead of only queuing up the last one.
let me know if you just want to wash your hands of this and i may be able to handle it then

the other thing is that with the new commands you can just make macros for them instead of always referring to `RunNewCommand` or something like that. constants for these new commands would also be very nice, at least a start to being able to add on from there

TixoRebel:
> super sorry for getting to this so late. would you mind throwing the repel that was used into the save so that it persists between reboots? at the end of the [`SAVE_MISC_DATA`](https://github.com/BluRosie/hg-engine/blob/main/include/save.h#L116) would be perfect, could then be accessed by `Sav2_Misc_get(SaveBlock2_get())->repelType` or something. may also be nice to have logic that queues up repels in order of max/super/normal instead of only queuing up the last one. let me know if you just want to wash your hands of this and i may be able to handle it then
> 
> the other thing is that with the new commands you can just make macros for them instead of always referring to `RunNewCommand` or something like that. constants for these new commands would also be very nice, at least a start to being able to add on from there

No worries, thanks for the feedback. I'll make the requested changes and update the PR hopefully this weekend.

Issue #293, opened 2024-07-20
Wrong interaction between Gigaton Hammer and Assault Vest


Aeliko:
Assault Vest should not block Gigaton Hammer
(Unconfirmed)

BluRosie:
appears fixed by bbf04e5a06001fd67999e87c1a2c387c49e47a1b as well

PR #296, opened 2024-07-21
Miscellaneous Changes

Aeliko:
# Moves
- [x] Implement Sparkling Aria
# Abilities
- [x] Implement Unnerve (#199 )
# Code
- [x] Move Parental Bond flag to somewhere else
- [x] Refactor Assault Vest check (indirectly related to #293 )

### DIFF
--- a/asm/include/battle_commands.inc
+++ b/asm/include/battle_commands.inc
@@ -1368,7 +1368,7 @@
     .long (((\arg_4 - .) / 4) - 1)
     .endm
 
-    .macro TurnOnAbilityActivatedFlag arg_0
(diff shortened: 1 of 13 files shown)


asm/include/battle_commands.inc
```diff
@@ -1368,4 +1368,9 @@
     .long (((\arg_4 - .) / 4) - 1)
     .endm
 
+    .macro TurnOnAbilityActivatedFlag arg_0
```
BluRosie, line 1371: maybe just `SetAbilityActivatedFlag`?  not all that vital a change

BluRosie (approved):
very small nitpick but otherwise looks good to me 😁

Issue #307, opened 2024-08-26
u-turn hitting a battler with a rocky helmet damages the mon after switching


BluRosie:
https://discord.com/channels/446824489045721090/1201213967389966378/1277580296073646182

PR #310, opened 2024-09-03
Fix for U-Turn bug

Mixone-FinallyHere:
Addresses #307 by moving the rocky helmet check for uturn to uturn fucntion

Additionally removes it from base check to avoid switched in mon getting hit

A note

Fixing this made me see that Jabocca berry is NOT checking the item protective pads to prevent damage, does this need adding?


BluRosie:
were you able to test this out?

BluRosie (approved):

Mixone-FinallyHere:
> were you able to test this out?

Yes, Mojo did

They confirmed it works

But I personally was not able to test

PR #311, opened 2024-09-03
Handle contact modern mecanics

Mixone-FinallyHere:
Handles contact with modern mecanics and leaves easy framework to add or remove item effects and abilities that can prevent contact

It includes also the stuff for PR #310 i edited that on main instead of a branch so its all together (sorry)

### DIFF
--- a/src/battle/other_battle_calculators.c
+++ b/src/battle/other_battle_calculators.c
@@ -2734,6 +2734,55 @@ BOOL CheckStrongWindsWeaken(struct BattleSystem *bw, struct BattleStruct *sp) {
     return FALSE;
 }
 
+/** 
+ * @brief checks if contact is being made, checking abilities and items
+ * @param bw battle work structure
+ * @param sp global battle structure
+ * @return TRUE/FALSE
+*/
+BOOL IsContactBeingMade(struct BattleSystem *bw, struct BattleStruct *sp) {
+
+    // Attacker abilities
+    if (GetBattlerAbility(sp, sp->attack_client) == ABILITY_LONG_REACH
+        // Kept in case people add their own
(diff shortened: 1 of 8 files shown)


BluRosie:
so...  we know that this covers the existing contact things in hg-engine, do we know about everything beyond this as well?  or are we thinking of tackle it as needed?

Mixone-FinallyHere:
> so... we know that this covers the existing contact things in hg-engine, do we know about everything beyond this as well? or are we thinking of tackle it as needed?

Im not SUPER familiar with like the most modern mecanics but between mojo and aero they mentioned all cases for contact that exist (PGloves + PPads and Long Reach)

The idea is that since the door is left open to easily add others, if in the future we see more its easy to simply append to this function I would say

TL;DR: As far as I know this tackles all possible use cases ASSUMING that neutralizing gas will be implemented at CheckAbility level

src/battle/other_battle_calculators.c
```diff
@@ -2734,6 +2734,55 @@ BOOL CheckStrongWindsWeaken(struct BattleSystem *bw, struct BattleStruct *sp) {
...
+        // Kept in case people add their own
+        // || HeldItemHoldEffectGet(sp, sp->defence_client) != OTHER_HOLD_EFFECT_THAT_PREVENTS_DEFENDER_CONTACT
+        ) {
+            return FALSE;
+    }
+
+    // Does the move make contact vanilla
+    if (IsContactBeingMade(bw, sp)) {
+        return TRUE;
+    }
```
BluRosie, line 2781: where is this linked into vanilla?  currently seems like it would recursively call itself and would likely stack overflow

BluRosie: should probably just be `sp->moveTbl[sp->current_move_index].flag & FLAG_CONTACT`

BluRosie (changes requested):

BluRosie (approved):

PR #312, opened 2024-09-04
End battle item exposure

Mixone-FinallyHere:
Exposes EndBattleItem handling for any purpose needed such as altering chances or adding other stuff (fetch pokeball ability thing or shuckle berry)

It includes also the stuff for PR #310  i edited that on main instead of a branch so its all together (sorry)


rom.ld
```diff
@@ -665,3 +665,9 @@ BattleContext_CheckMoveHealBlocked = 0x02252DF8 | 1;
 CanUseItemOnPokemon = 0x0208FD9C | 1;
 
 BagApp_GetSaveRoamers = 0x021F992C | 1;
+
+/* Stuff for Battle End Item */
+sPickupTable1 = 0x0226C404 | 1;
+sPickupTable2 = 0x0226C30C | 1;
+sPickupWeightTable = 0x0226CA4C | 1;
+sHoneyGatherChanceTable = 0x0226CA58 | 1;
```
BluRosie, line 673: data should not have `| 1` as that is for thumb code and will mess up loads

PR #321, opened 2024-09-11
Modernise Metronome, Nature Power message

Aeliko:


data/text/197.txt
```diff
@@ -488,7 +488,7 @@ The opposing {STRVAR_1 1, 0, 0} became\nthe center of attention!
 {STRVAR_1 1, 0, 0} began\ncharging power!
 The wild {STRVAR_1 1, 0, 0} began\ncharging power!
 The opposing {STRVAR_1 1, 0, 0} began\ncharging power!
-{STRVAR_1 6, 0, 0} turned into\n{STRVAR_1 6, 1, 0}!
+Nature Power turned into\n{STRVAR_1 6, 1, 0}!
```
BluRosie, line 491: what was the issue here?

BluRosie (approved):

data/text/197.txt
Aeliko: That's just how it is in later gens, they hard coded it similar to Reflect, etc.

Issue #308, opened 2024-08-26
spikes/toxic spikes should not proc on a battler holding an air balloon


BluRosie:
https://discord.com/channels/1221517397874446557/1227093922954608691/1277225478767251478

BluRosie:
done by this line https://github.com/BluRosie/hg-engine/blob/d6c63d8007425f5e205a90a89a73a2d807e25361/data/battle_scripts/subscripts/subscript_0099_HAZARDS_CHECK.s#L34

PR #322, opened 2024-09-11
Sticky Web

Koekenpann:
Few notes:

- The Defiant and Competitive case is probably not the best solution as it doesn't cover all the switch-in cases like Intimidate.
- However there is no longer a need for an edge-case where a Court Changed Sticky Web doesn't activate Defiant or Competitive. I have verified this on cartridge which I can show on Discord.
- It should have (some) foundation for Mirror Armor, as it stores the user, users side and the existence of Sticky Web as Aero suggested on Discord.
- [Contrary raises its Speed.](https://i.imgur.com/B5FuZ1v.gif)
- Defog (both sides) and Rapid Spin (user's side) remove the Sticky Web from the field.

melonDS_20240801-195931210

melonDS_20240801-195950447

Also fixes:
- Heavy-Duty Boots didn't protect Flying-types/Levitate Pokémon from Stealth Rock (I think).
- If Rapid Spin fainted something, then the switched-in Pokémon triggered a stat drop it would re-run the RapidSpin battlecommand.
- If a Comatose Pokémon switches in on Toxic Spikes + another entry hazard, it would not be affected by any entry hazard.
- Seems like #308 was already fixed before these changes.

Hope I covered everything, if not let me know!


src/battle/battle_script_commands.c
```diff
@@ -3184,7 +3187,43 @@ BOOL BtlCmd_RapidSpin(void *bw, struct BattleStruct *sp)
...
     IncrementBattleScriptPtr(sp, 1);
+    int adrs = read_battle_script_param(sp);
+    int side = IsClientEnemy(bw, sp->defence_client);
+
+    if (sp->side_condition[side] & SIDE_STATUS_STICKY_WEB) {
+        sp->oneSelfFlag[sp->attack_client].no_pressure_flag = 1;
+        IncrementBattleScriptPtr(sp, adrs);
+    } else {
+        sp->side_condition[side] |= SIDE_STATUS_STICKY_WEB;
+        sp->scw[side].stickyWebBattlerId = sp->attack_client;                      // For Mirror Armor in doubles
```
Aeliko, line 3224: If the Pokémon with this Ability is affected by [Sticky Web](https://bulbapedia.bulbagarden.net/wiki/Sticky_Web_(move)), the effect is reflected back to the Pokémon which set it up. If Pokémon which set up Sticky Web is not on the field, no Pokémon have their Speed lowered.

`sp->attack_client` is the position of the active battlers, would need to somehow handle this in however way you think is good

data/battle_scripts/subscripts/subscript_0012_UPDATE_STAT_STAGE.s
```diff
@@ -14,6 +14,7 @@ _010:
     Wait 
 
 _018:
+    CompareVarToValue OPCODE_FLAG_NOT, BSCRIPT_VAR_BATTLE_STATUS, BATTLE_STATUS_MOVE_SUCCESSFUL, _023
```
BluRosie, line 17: oh shucks, was rapid spin always raising speed?  that's funny
should really do rapid spin properly at some point 😂

src/battle/battle_script_commands.c
BluRosie: wait so...  mirror armor switch in will put the sticky web back on the other side of the field *IF* the mon that set it up is still there?

Aeliko: The stat drop itself, not the hazard.

BluRosie: oh thank goodness
so
`sp->scw[side].stickyWebBattlerId` is default 0, client + 1 if active, and we clear it with the other battle flags in [`ClearBattleMonFlags`](https://github.com/BluRosie/hg-engine/blob/3129587bfb4f0b6b3461aaba3f0feb435c17b8dc/src/battle/battle_pokemon.c#L992) maybe?  default `MAX_CLIENT` and `client` if active?

Aeliko: Sounds good to me!

BluRosie (approved):

PR #326, opened 2024-09-25
Ai can use trick

kumakuma1:
https://github.com/pret/pokeheartgold/blob/fa9d1977b4bf4f7439cb02ebb31c1dafd766d49c/src/battle/battle_command.c#L3816

- config to allow AI to use trick in battles
- modernize tricked items
-Thief, Bestow and Switcheroo could be similar, or could use the helper function


kumakuma1:
upsi

BluRosie (approved):

PR #336, opened 2024-10-19
Convert flags.s from hex to decimal

FrozenMetroid:
Ew hexadecimal numbers


BluRosie (approved):
noooo my emerald-specific flag names

PR #349, opened 2024-10-29
Disable items in trainer battle

kumakuma1:
- add option to disable items in Trainer Battles
- AI can't use items either
- When you click on Bag during the fight:
Screenshot 2024-10-29 162404

In case there is an easier way, delete this.


BluRosie (approved):
nice 😁

PR #350, opened 2024-11-01
Npc trade

kumakuma1:
- expose _CreateTradeMon to edit the trades Pokemon.
- decrompress overlay 23
- _CreateTradeMon can be adapted to trade any mon with any settings
- Here are two examples:
- I commented out the ShinyCheck
Unbenannt
image


BluRosie (commented):
just need to then put the table in to allow expansion?

kumakuma1:
> just need to then put the table in to allow expansion?
> 
> also the shiny check was just there to make sure that the mon isn't accidentally shiny with the present pid and otid

You need at least NPCTradeApp_Init from 
https://github.com/pret/pokeheartgold/blob/master/src/npc_trade.c#L25

I just edited the function with a big switch statement for all trades.

PR #358, opened 2024-11-14
Converts Flags.s From Hex to Decimal and Update Scriptmacros.s With More Constants

FrozenMetroid:


BluRosie (commented):
make the tabs spaces 💢

PR #370, opened 2024-12-28
Very Fast HP Bar

blackisme1:
It makes the HP bar very fast

https://github.com/user-attachments/assets/ea71cc7b-b90a-468c-a5e3-356ffe47ca1a

More specifically, it makes every HP bar act like it has less than 150 max hp, resulting in HP above 150 to use bigger divisions of the bar


bytereplacement
```diff
@@ -490,3 +490,9 @@ arm9 080550D4 2C
 0012 0223FBD0 81
 0012 0223FBD4 81
 0012 0223FBDC 81
+
+# change ble to b for very fast HP bar
```
Aeliko, line 494: Add config to toggle fast or slow HP bar

blackisme1: alright

blackisme1:
Game freezes when leveling up, will put this pr as WIP until fixed

blackisme1:
https://github.com/user-attachments/assets/51dd553b-2236-4d04-95b2-2590348adb1b

i fixed it

PR #402, opened 2025-04-05
Modifying snow to be in overworld

agathaGengar:


src/battle/other_battle_calculators.c
```diff
@@ -738,13 +738,6 @@ u8 LONG_CALL CalcSpeed(void *bw, struct BattleStruct *sp, int client1, int clien
         {
             speed1 *= 2;
         }
-        if (((ability2 == ABILITY_SWIFT_SWIM) && (sp->field_condition & WEATHER_RAIN_ANY))
-         || ((ability2 == ABILITY_CHLOROPHYLL) && (sp->field_condition & WEATHER_SUNNY_ANY))
-         || ((ability2 == ABILITY_SAND_RUSH) && (sp->field_condition & WEATHER_SANDSTORM_ANY))
-         || ((ability2 == ABILITY_SLUSH_RUSH) && (sp->field_condition & (WEATHER_HAIL_ANY | WEATHER_SNOW_ANY))))
-        {
-            speed2 *= 2;
-        }
```
Aeliko, line 747: Why is this removed?

Aeliko (changes requested):

src/battle/other_battle_calculators.c
agathaGengar: Mistake on my part, it doesn't need to change. I thought it was duplicate, but they have ability 1 and ability 2 as different blocks.

PR #421, opened 2025-05-16
Q412 Damage Calculator 2 Electric Boogaloo

### LINKED ISSUE #184: damage calculator parity with later generations
see <https://github.com/lhearachel/plat-engine/blob/main/src/battle/calc.c> for implementation details

Aeliko:
# Summary
What's up guys, I'm back with another banger 🔥

Closes #184 

I cannot directly link my projects to this PR, so here are the relevant ones:
[Damage](https://github.com/users/Aeliko/projects/1?pane=issue&itemId=100423039)

# Details
The damage calculation uses Q4.12 format numbers. The implementation (`qmath.s`, `q412.h`) is provided by Lhea.

Random damage modifier is moved inside damage calculation, so the original function is emptied.

Note that the functions have side effects, which should be fine?

What's that? You want to know why is there that weird define? That one? Yes, yes, that one is crucial, it is imperative that it is in the function. DO NOT REMOVE IT!!!


BluRosie:
for some reason i don't recall off the top of my head, i don't like pragma.  is it a gcc thing and we're completely fine using it?

Aeliko:
Admittedly I just copied from Stack Overflow and it made the compiler shut up about those warnings. Should not be risky?

src/battle/battle_pokemon.c
```diff
@@ -1223,3 +1223,7 @@ BOOL LONG_CALL DoesSideHave2Battlers(void *bw, u32 client)
     }
     return FALSE;
 }
+
+BOOL LONG_CALL ClientBelongsToPlayer(struct BattleSystem *bsys, int client) {
+    return BattleWork_GetTrainerIndex(bsys, client) == 0;
+}
```
BluRosie, line 1232: have you made sure that this check is still true for i.e. partner battles?

src/battle/other_battle_calculators.c
```diff
@@ -422,15 +422,18 @@ BOOL CalcAccuracy(void *bw, struct BattleStruct *sp, int attacker, int defender,
...
     u8 move_type UNUSED; // unused but will be needed
     u8 move_split;
     u16 atk_ability = GetBattlerAbility(sp, attacker);
     int i;
     int maxBattlers = BattleWorkClientSetMaxGet(bw);
     int accuracyModifier = UQ412__1_0;
 
+    if (I_AM_TERAPAGOS_AND_I_NEED_TO_KO_CARMINES_SINISTCHA(bw, sp, attacker)) {
+        return FALSE;
+    }
```
BluRosie, line 456: isn't this an individual game-defined thing and not something that could even be generic?

i borderline feel this way about friendship effects as well

include/config.h
```diff
@@ -77,6 +77,15 @@
...
 //#define UNCAP_CANDIES_FROM_LEVEL_CAP
 
+// System flags that need to be enabled for the player to use the gimmick. If you want to change them, remember to change them in flags.s as well for consistency
+#define FLAG_MEGA_EVOLUTION_ENABLED 2518
+#define FLAG_Z_MOVE_ENABLED 2519
+#define FLAG_DYNAMAX_ENABLED 2520
+#define FLAG_TERASTALIZATION_ENABLED 2521
+
+// Uncomment this when you want to enable all gimmicks while testing
+// #define DEBUG_ENABLE_ALL_GIMMICKS
```
BluRosie: put in debug.h?

src/battle/other_battle_calculators.c
```diff
@@ -1624,32 +1641,6 @@ int LONG_CALL ServerDoTypeCalcMod(void *bw UNUSED, struct BattleStruct *sp, int
...
-                    damage = damage * (100 + atk_a) / 100;
-                }
-            }
-            if ((flag[0] & MOVE_STATUS_FLAG_NOT_VERY_EFFECTIVE) && (base_power))
-            {
-                if (GetBattlerAbility(sp, attack_client) == ABILITY_TINTED_LENS)
-                {
-                    damage *= 2;
-                }
-            }
```
BluRosie, line 1681: what does all this do now?  just moved to damage calculator?

src/battle/other_battle_calculators.c
```diff
@@ -2303,20 +2293,20 @@ BOOL LONG_CALL BattleSystem_CheckMoveEffect(void *bw, struct BattleStruct *sp, i
...
         }
     }
 
     if (sp->battlemon[battlerIdTarget].effect_of_moves & MOVE_EFFECT_FLAG_MINIMIZED
         && !sp->battlemon[battlerIdTarget].is_currently_dynamaxed
         && IsMoveInMinimizeVulnerabilityMovesList(move)) {
-        sp->waza_status_flag &= ~MOVE_STATUS_FLAG_MISS;
-        return TRUE;
+            sp->waza_status_flag &= ~MOVE_STATUS_FLAG_MISS;
+            return FALSE;
```
BluRosie, line 2346: why the extra tabs here?

src/individual/CalcBaseDamage.c
```diff
@@ -126,55 +142,116 @@ static const u16 SharpnessMovesTable[] = {
...
+    {   250,     40}, //  10- 25 kg ->  40 bp
+    {   500,     60}, //  25- 50 kg ->  60 bp
+    {  1000,     80}, //  50-100 kg ->  80 bp
+    {  2000,    100}, // 100-200 kg -> 100 bp
+    {0xFFFF, 0xFFFF},
+};
 
 
+// TODO
+// Why is it offset if I use the original definition??????????????
```
BluRosie, line 157: what's the problem here?

src/individual/CalcBaseDamage.c
```diff
@@ -126,55 +142,116 @@ static const u16 SharpnessMovesTable[] = {
...
+        return 9999;
     }
 
-    defense = BattlePokemonParamGet(sp, defender, BATTLE_MON_DATA_DEF, NULL);
-    sp_attack = BattlePokemonParamGet(sp, attacker, BATTLE_MON_DATA_SPATK, NULL);
-    sp_defense = BattlePokemonParamGet(sp, defender, BATTLE_MON_DATA_SPDEF, NULL);
+    // https://web.archive.org/web/20241226231016/https://www.trainertower.com/dawoblefets-damage-dissertation/
+
+    // switch (moveno) {
+    //     // handle body press - attack is derived from defense
```
BluRosie, line 217: do we no longer handle body press then?

src/individual/CalcBaseDamage.c
```diff
@@ -225,836 +310,1421 @@ int CalcBaseDamage(void *bw, struct BattleStruct *sp, int moveno, u32 side_cond,
...
+        break;
+    case MOVE_PAYBACK:
+        // TODO: Check correctness
+        if (sp->playerActions[defender][0] == CONTROLLER_COMMAND_40) {
+            movepower *= 2;
+        }
+        break;
+    case MOVE_PURSUIT:
+        // TODO: Handle this massive headache later
+        movepower = movepower * sp->damage_value / 10;
```
BluRosie, line 521: does pursuit now just always multiply or what?  how is damage_value set here?

src/battle/battle_calc_damage.c
```diff
@@ -96,57 +97,603 @@ u16 LONG_CALL GetBattleMonItem(struct BattleStruct *sp, int client_no)
...
+#ifdef DEBUG_DAMAGE_CALC
+        debug_printf("[CalcBaseDamage] parental_bond_flag: %d\n", sp->oneTurnFlag[attacker].parental_bond_flag);
+#endif
+        switch (sp->oneTurnFlag[attacker].parental_bond_flag) {
+            case 1:
+                sp->oneTurnFlag[attacker].parental_bond_flag++;
+                sp->oneTurnFlag[attacker].parental_bond_is_active = TRUE;  // after first hit, set this flag just in case the ability is nullified after the first one
+                break;
+            case 2:
+                debug_printf("[CalcBaseDamage] parental_bond_flag is 2, damage reduction\n");
```
BluRosie, line 175: no debug wraps

src/battle/battle_calc_damage.c
```diff
@@ -96,57 +97,603 @@ u16 LONG_CALL GetBattleMonItem(struct BattleStruct *sp, int client_no)
...
+        if ((attackerAbility == ABILITY_SNIPER) && (sp->critical > 1)) {
+            finalModifier = QMul_RoundUp(finalModifier, UQ412__1_5);
+#ifdef DEBUG_DAMAGE_CALC
+            debug_printf("\n=================\n");
+            debug_printf("[CalcBaseDamage] 6.9.3 Sniper\n");
+            debug_printf("[CalcBaseDamage] finalModifier: %d\n", finalModifier);
+#endif
+        }
+
+        if ((sp->rawSpeedNonRNGClientOrder[i] == BATTLER_ALLY(defender)) && (GetBattlerAbility(sp, BATTLER_ALLY(defender)) == ABILITY_FLUFFY)) {
```
BluRosie: why is the fluffy check on `BATTLER_ALLY(defender)` instead of just defender?

src/individual/CalcBaseDamage.c
```diff
@@ -225,836 +310,1421 @@ int CalcBaseDamage(void *bw, struct BattleStruct *sp, int moveno, u32 side_cond,
...
+    // TODO: Check if there are any moves not ported yet 
+    // if (pow == 0) {
+    //     debug_printf("First case\n");
+    //     movepower = sp->moveTbl[moveno].power;
+    // } else {
+    //     debug_printf("Second case\n");
+    //     movepower = pow;
+    // }
+
+    switch (moveno) {
```
BluRosie, line 328: are we good to swap to per-move instead of per-effect?  i understand the benefits in terms of thinking about things and understanding, but this splits from being able to just specify the effect in movedata and makes for an overall inflexible implementation for rom hacks that want custom moves--"i want a fairy type gyro ball"

src/individual/CalcBaseDamage.c
```diff
@@ -225,836 +310,1421 @@ int CalcBaseDamage(void *bw, struct BattleStruct *sp, int moveno, u32 side_cond,
...
+        movepower = (255 - AttackingMon.happiness) * 10 / 25;
+        break;
+    // Counter-based
+    case MOVE_FURY_CUTTER:
+        for (int n = 0; n < sp->battlemon[attacker].moveeffect.furyCutterCount; n++) {
+            movepower *= 2;
+        }
+        break;
+    case MOVE_ROLLOUT:
+    case MOVE_ICE_BALL:
```
BluRosie, line 450: would also just be one case if we kept move effects

src/individual/CalcBaseDamage.c
```diff
@@ -225,836 +310,1421 @@ int CalcBaseDamage(void *bw, struct BattleStruct *sp, int moveno, u32 side_cond,
...
+            if ((AttackingMon.item_held_effect == HOLD_EFFECT_CHOICE_ATK) && (movesplit == SPLIT_PHYSICAL)) {
+                attackModifier = QMul_RoundUp(attackModifier, UQ412__1_5);
+            }
+            
+            // handle Choice Specs
+            if ((AttackingMon.item_held_effect == HOLD_EFFECT_CHOICE_SPATK) && (movesplit == SPLIT_SPECIAL)) {
+                attackModifier = QMul_RoundUp(attackModifier, UQ412__1_5);
+            }
+    
+            // TODO: it’s not a Ditto/Smeargle/Mew Transformed into the species <- how to do this?
```
BluRosie: should be handled by just checking the species?

src/individual/CalcBaseDamage.c
```diff
@@ -225,836 +310,1421 @@ int CalcBaseDamage(void *bw, struct BattleStruct *sp, int moveno, u32 side_cond,
...
+            && (movesplit == SPLIT_PHYSICAL)) {
+                defenseModifier = QMul_RoundUp(defenseModifier, UQ412__2_0);
+            }
+        }
+    }
+
+    // Apply the chained modifier to the starting defense. That is, multiply the starting defense by the chained defense modifiers, divide by 4096, and pokeRound the result. If the current defense would now be less than 1, make it 1. Finally, if the defense is greater than 65,535, make it the defense modulo 65,536 (defense % 65536). If the defense stat is 0 because of this modifier, the result of base damage will always be 2.
+    calculatedDefense = QMul_RoundDown(calculatedDefense, defenseModifier);
+    calculatedDefense = calculatedDefense < 1 ? 1 : calculatedDefense;
+    calculatedDefense = calculatedDefense % 65536;
```
BluRosie, line 1671: general question for all of the modulo operations--this can still result in a 0 calculatedDefense.  does that mimic future implementations?

BluRosie (commented):
few things here.  i think we also need to take stock of every ability/item/move that is handled and make sure we're not leaving things out.  i may or may not get to this soon before merging, or may just do it after merging--but it has to be done soon.  if this didn't fix other bugs i'd like to get it done before merging.

src/battle/battle_calc_damage.c
Aeliko: You're right

src/battle/other_battle_calculators.c
Aeliko: I intend to implement all of them, like Totem battles, SwSh/SV raids and have the logic check the battle type (which would need new types defined for them)

src/individual/CalcBaseDamage.c
Aeliko: It needs to have a higher move power if the defender is switching out, but we don't have code checking for that yet. Instead we just use the original logic in the subscript for now

src/individual/CalcBaseDamage.c
Aeliko: I guess so

I think we should do a move effect naming refactor then? To make it easier to conceptualise which underlying move it is referring to

src/individual/CalcBaseDamage.c
Aeliko: I think vanilla Gen 4 doesn't store the pre-transformed species?

src/individual/CalcBaseDamage.c
Aeliko: Yes, there should be logic guarding such cases

src/individual/CalcBaseDamage.c
Aeliko: I cannot use the stat boost table in the other file. I had the duplicate the definition here for some reason otherwise it fetches the wrong values

src/battle/other_battle_calculators.c
Aeliko: Yes

src/individual/CalcBaseDamage.c
BluRosie: oh wow, pretty funny.  should be able to check `condition2 & STATUS2_TRANSFORMED` though

src/individual/CalcBaseDamage.c
BluRosie: will take a look once merged

PR #423, opened 2025-05-18
Add modern cries (Gen 1 - 8)

Aeliko:
Maybe they need to be a little bit quieter? Other than that should be fine?


Aeliko:
Should be Meltan (808.ogg in the source repo, 858.wav in hg-engine) and Toedscruel (949.ogg in the source repo, 999.wav), and the misc forms (just haven't got the time to convert those yet). Don't know why these 2 files in particular have issues.

PR #427, opened 2025-05-19
get rid of remainder of _hook constructs

ZeroARND:


BluRosie:
> 

heads up--this commit is a little unstable and was removed from main for that reason.  `Bag_HasItem` isn't compatible with the new hook format and causes crashes.  will be fixed really soon though!

PR #428, opened 2025-05-21
Add Oricorio form changes

BusyBidoof:
I was experimenting a bit with form changes, and implemented this as an example. 
It isn't much, but I hope each bit helps in a project like this.


BluRosie (approved):
looks good!  embarrassing on my part for the mint function 😂

Issue #418, opened 2025-05-10
Issue with Stat Change preventing Abilities


SteeledKnight:
I've only been able to test with Keen Eye so far, but I suspect more are affected. When Keen Eye activates, nothing happens, like your turn is skipped. Since the stat isn't lowered, it looks like the check for Keen Eye in other_battle_calculators.c is catching it, but 'subscript_0396_ACCURACY_NOT_LOWERED.s' seemingly isn't called or not call correctly, since nothing in that script happens. I've tested this with my current version of my romhack, then also tested it with a cleanly pulled version of hg-engine with a vanilla rom, built with no changes made to config.s, config.h, or anywhere else. The behavior is the same.

SteeledKnight:
After further testing, ALL abilities preventing stat drops were affected. It looks like the function for checking if stats were as low as possible was piggybacked off of for the ability blocking checks. I cleaned up the code so behavior between the two functions was more consistent. I believe that order of logic was overwriting the appropriate subscript's place in the battle sequence.

PR #431, opened 2025-05-27
Fix abilities that prevent stats from being lowered

SteeledKnight:
Addresses #418

Updated the handling of loading the proper subscript for providing messaging about the ability preventing stat changes.


BluRosie (commented):
so it just unnecessarily loaded the subscript itself?  or what's the deal?  looks good though 😁

SteeledKnight:
More like it was queued up in one place then overwritten/removed from the sequence before executing in another, from what I could tell. I think there was just an oversight in the order of operations between calling a function and returning from it. I just partially rewrote the functions to simplify that hand-off. Sorry if my explanation is lacking. I'm still trying to wrap my head around this codebase 😅
However, it does look like I need to rebase before this gets merged, or it's going to overwrite that recent change to definitions in battle.h. I should be able to address that tomorrow

BluRosie (approved):

PR #436, opened 2025-06-04
Electromorphosis, Wind Power & Wind Rider

Briocheee:
Missing Tailwind interaction
More tests to Wind Power and Electromorphosis (was giving 2x damage against one pokemon, but there's room for more tests)


data/text/197.txt
BluRosie, line 1: need an opposing string as well

BluRosie (approved):

PR #443, opened 2025-06-12
Item index reordering

LucaPrevi0o:
# Implement global dynamic offsets
Allow subsets of items (Balls, Berries, Plates...) to be re-ordered as *consecutive item indices*, with the help of dynamic offsets that automatically update the index of the other items when new/custom items are added.

This helps organize item indices and improves item expansion, as the index offset calculation is no longer bound to the current number of items, but rather to the number of different items in a specific subset.

> **DISCLAIMER**
> Actual values for item indices have not been changed. Only reference to global offset has been re-calculated based on relative start index.

## Changes
* Separate header files have been inserted in the [`include/constants`](https://github.com/BluRosie/hg-engine/tree/main/include/constants) folder, under the new subfolder `items`
* Relative offset for newer items (pointing to `ITEM_PIXIE_PLATE`) have been updated to new `ITEM_START` global offset for each category
* Included `#define` macros for `IS_ITEM` and `NUM_ITEMS` in every category

## Future developments
* Update compilation phase to include new offset system
* Include missing item subsets
* Possibly rework Pokemon/Pokedex variant form system in the same way


Aeliko:
We actually plan to add all new items and order them according to how it is in the latest mainline games. Custom items would go after the vanilla items. How would this affect your PR, if any?

LucaPrevi0o:
just to clarify - are newer items (like newer berries/balls etc) put after everything *even* in mainline games? aren't they reworked every time? like, is pixie plate item 584 even if other plates are 280-ish?

because my point is, let's aggregate similar items together (like plates and others), supposing this is what happens in mainline games (if not, I would be very surprised, ngl), and then custom items can come after vanilla ones like before - with the exception that new balls/berries/etc that work similarly to vanilla items can automatically be put in the same subset of items

BluRosie:
what are we thinking?

BluRosie:
superseded by #469

PR #444, opened 2025-06-12
fixed ITEM_EXPANSION_START as offset for newer items

LucaPrevi0o:
all items in `items.h` still used `ITEM_PIXIE_PLATE` as offset for constant definition

in this way, `ITEM_EXPANSION_START` is correctly implemented


include/constants/item.h
```diff
@@ -306,7 +306,7 @@
 #define ITEM_BIG_ROOT        296
 #define ITEM_CHOICE_SPECS    297
 
-#define IS_ITEM_ARCEUS_PLATE(item) ((item >= ITEM_FLAME_PLATE && item <= ITEM_IRON_PLATE) || (item == ITEM_PIXIE_PLATE))
+#define IS_ITEM_ARCEUS_PLATE(item) ((item >= ITEM_FLAME_PLATE && item <= ITEM_IRON_PLATE) || (item == ITEM_EXPANSION_START))
```
BluRosie, line 309: i assume this was accidentally caught in a search and replace?

BluRosie (changes requested):
would you also want to move `ITEM_EXPANSION_START` to be the start of the mega stones?

include/constants/item.h
LucaPrevi0o: it's actually the same tbh, maybe it's a little more clear if you leave ITEM_PIXIE_PLATE because you reference directly every plate, so if there's need to put anything before the pixie plate it will be unaffected... up to you if you want to change

BluRosie:
superseded by #469

PR #445, opened 2025-06-12
add trainer validation script

zebben:
Let me know if you think there'd be a better way to integrate/achieve this. Pretty simple just adding validation to the regular `make` process to help with improper trainer setup.

Some examples of the output below. Could look at capitalizing errors or something if we want to improve visibility

nummons not matching actual party size

```
error: Josh (id: 647, nummons: 3) has 2 pokémon defined
make: *** [Makefile:266: all] Error 1
```

fields out of order

```
error: Josh (id: 647 mon 0: field order is incorrect.
  found order: ['ivs', 'abilityslot', 'pokemon', 'level', 'ballseal']
  expected order (if present): ['ivs', 'abilityslot', 'level', 'pokemon', 'item', 'move1', 'move2', 'move3', 'move4', 'ability', 'setivs', 'setevs', 'nature', 'shinylock', 'additionalflags', 'ballseal']
make: *** [Makefile:266: all] Error 1
```

missing `trainermontype` flag

```
error: Misty (id: 721, flags: TRAINER_DATA_TYPE_MOVES) does not have the TRAINER_DATA_TYPE_ITEMS flag but some party pokémon have an item defined
make: *** [Makefile:266: all] Error 1
```

missing mon values based on set flags
```
error: Misty (id: 721, flags: TRAINER_DATA_TYPE_ITEMS | TRAINER_DATA_TYPE_MOVES) has TRAINER_DATA_TYPE_ITEMS flag but not all party pokémon have an item defined
make: *** [Makefile:266: all] Error 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -263,6 +263,7 @@ all: $(TOOLS) $(OUTPUT) $(OVERLAY_OUTPUTS)
 	@echo "$(ROMNAME) Decompression successful!!"
 	$(NARCHIVE) extract $(FILESYS)/a/0/2/8 -o $(BUILD)/a028/ -nf
 	$(PYTHON) scripts/make.py
+	$(PYTHON) scripts/validate_trainers_s.py $(TRAINERDATA_DEPENDENCIES)
 	$(MAKE) move_narc
 	$(ARMIPS) armips/global.s
 	$(NARCHIVE) create $(FILESYS)/a/0/2/8 $(BUILD)/a028/ -nf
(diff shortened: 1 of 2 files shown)


BluRosie:
things look good.  how does it handle `additionalflags` and such?

Makefile
BluRosie, line 1: might just need to do it when the trainers are being built, i.e. right around here https://github.com/BluRosie/hg-engine/blob/main/narcs.mk#L233

BluRosie (approved):
looks good to me.  will merge and do some testing

Makefile
zebben: ah yeah that'd make sense. let me know if you want me to PR that or any other changes needed

PR #454, opened 2025-06-25
Protect moves

Koekenpann:
# Protect moves
Add the new Protect moves added in Gens 5 to 9.

1. Move failure cases are now handled in ``BattleController_BeforeMove.c`` instead of ``TryProtection`` battle command.
2. ``TryProtection`` will then set the flag on the Pokémon that should be protecting.
3. ``BattleController_BeforeMove.c`` will then check if the "Protecting" flag is active, then check the other conditions (the scope) to be protected against the move (eg. King's Shield can only protect against attacking moves).
4. If an additional effect must be triggered upon contact, ``CheckProtectContactMoves`` battle command will load the proper subscript.

## Modernizations
* Successive Protect turns (partially, up to 4 consecutive Protects)

Should go up to 6: https://bulbapedia.bulbagarden.net/wiki/Protection#Generation_VI

## Fixes
* Fix some unprotectable move flags as of: https://bulbapedia.bulbagarden.net/wiki/Protection#Unprotectable_moves
* Beast Boost highest stat grabber overflowing if stat is higher than 255 (small fix, might as well)

## New Ability Implementations
* Unseen Fist

## New Moves Implementations
* Wide Guard
* Quick Guard
* Mat Block
* Crafty Shield
* King's Shield
* Spiky Shield
* Baneful Bunker
* Max Guard (partially)
* Obstruct
* Silk Trap
* Burning Bulwark

## Example
melonDS_20250625-221007331


BluRosie (approved):
changes look perfect to me!

PR #458, opened 2025-07-03
Learnset overhaul

zebben:
### Summary

- Overhauls mon learnsets to improve user experience and accuracy to the mainline games
- Dynamic expansion for TM/HM/TR (machine) moves, Egg Moves, Level Up Moves, and Tutor Moves
- Unifies learnsets so users don't have to modify multiple files for a single mon's learnsets
- File-order based inheritance allows for canonical moveset definitions of older generations and the final unified JSON file is dumped to `build/learnsets.json` for debugging purposes
- Users can add new files per mon if they wish to, or just use the `custom` JSON file for all of their modifications

Format example:
```JSON
"SPECIES_BULBASAUR": {
    "LevelMoves": [
      {
        "Level": 1,
        "Move": "MOVE_TACKLE"
      },
      {
        "Level": 1,
        "Move": "MOVE_GROWL"
      },
      {
        "Level": 3,
        "Move": "MOVE_GROWL"
      },
      {
        "Level": 5,
        "Move": "MOVE_VINE_WHIP"
      },
      {
        "Level": 8,
        "Move": "MOVE_LEECH_SEED"
      }
    ],
    "MachineMoves": [
      "MOVE_ATTRACT",
      "MOVE_BULLET_SEED",
      "MOVE_CAPTIVATE",
      "MOVE_CUT",
      "MOVE_DOUBLE_TEAM",
      "MOVE_ENDURE",
      "MOVE_ENERGY_BALL"
    ],
    "EggMoves": [
      "MOVE_LIGHT_SCREEN",
      "MOVE_SKULL_BASH",
      "MOVE_SLUDGE"
    ],
    "TutorMoves": [
      "MOVE_FURY_CUTTER",
      "MOVE_HEADBUTT",
      "MOVE_KNOCK_OFF",
      "MOVE_MUD_SLAP",
      "MOVE_SEED_BOMB"
    ]
  }
```

### Machine Moves
- Allow for TRs and brand new TMs/HMs to be added (not just modifying existing)
- Customized rendering for Machine Moves via configuration. This allows for better distinction between TMs / TRs / HMs
image

### Egg Moves &  Level Up Moves
- Not much is different here other than dynamic expansion

### Tutor Moves
- Users can now add more moves to the tutor move pool

#### Egg/Levelup
Inheritance takes the latest generation's available learnset as the explicit value for the output
#### Tutor/Machine
Inheritance is cumulative across consumed files, meaning a large list of all unique entries is output (flag in config.h to enable/disable behavior) but moves must be added to the appropriate C arrays to be effective in-game

### Limitations
- 255 move max for Tutor Moves. Would require some pretty significant overhauls of non-decomped code.I think this is a good middle ground for 99% of hacks
- 64 Level Up Moves. This is imposed because of the move relearner. Might be able to expand this further though without too much difficulty

### Migration
- `custom.json` is pre-filled with existing repo contents on main
- Must have previously built on old branch and maintained the `build` dir from that
- Run `python3 scripts/migrate_learnsets.py` to migrate to the json format from the s/txt
- Script prints outs Machine Moves and Tutor Moves which were modified by the user with instructions to copy to the appropriate C arrays


src/bag.c
```diff
@@ -444,3 +472,41 @@ BOOL IsPlayerOnLadder(void)
...
+    u16 labelId = args[1];
+
+    int numDigits = 3;
+    int icon = BAG_TM_ICON;
+
+    if (IS_ITEM_HM(itemId)) {
+        icon = BAG_HM_ICON;
+        numDigits = 2;
+    } else if (IS_ITEM_TR(itemId)) {
+        icon = BAG_TR_ICON;
```
Aeliko, line 516: 2 digits for TRs

src/pokemon.c
```diff
@@ -1237,42 +1237,25 @@ BOOL LONG_CALL Party_TryResetShaymin(struct Party *party, int min_max, const str
...
-        }
-        dest[i] = kowaza_list[offset + i];
-        n++;
+u8 LONG_CALL LoadEggMoves(struct PartyPokemon *pokemon, u16 *dest) {
+    debug_printf("[LoadEggMoves] species: %d, form: %d\n", GetMonData(pokemon, MON_DATA_SPECIES, NULL), GetMonData(pokemon, MON_DATA_FORM, NULL));
+    u16 species = PokeOtherFormMonsNoGet(GetMonData(pokemon, MON_DATA_SPECIES, NULL), GetMonData(pokemon, MON_DATA_FORM, NULL));
+    debug_printf("[LoadEggMoves] specieswithform: %d\n", species);
+
+    ArchiveDataLoadOfs(dest, ARC_EGG_MOVES, 0, species * EGG_MOVES_PER_MON * sizeof(u16), EGG_MOVES_PER_MON * sizeof(u16));
+    debug_printf("[LoadEggMoves] loaded.\n");
```
Aeliko: Surround with debug defines

zebben: I'll probably just remove once I've finished verifying unless you want wrapped debugs throughout

data/graphics/bag/95.png
zebben, line 1: the images here have strange colors. this is because I had to do a lot of fiddling to get the "TM"/"TR" part to match up to the HM icon's palette. I was unable to find the associated palette as it wasn't in a015 with the image. this is also why I'm not copying the palettes to the a015 for these

there is probably a way to make it look nice in the repo while maintaining the indexes but I honestly kind of lost patience with it

include/bag.h
```diff
@@ -48,6 +48,12 @@ typedef struct BagData {
     u16 registeredItems[2];                        // IDs of registered key items
 } BAG_DATA;
 
+typedef enum TMHMPocketSortPrecedence {
```
zebben, line 51: this could be renamed but wasn't sure if I should rename the pocket yet.

data/codetables.mk
```diff
@@ -68,3 +68,49 @@ $(FORMREVERSION_BIN): $(FORMREVERSION_DEPENDENCIES)
...
 
 NARC_FILES += $(FORMREVERSION_BIN)
+
+
+MACHINELEARNSET_TARGET := $(BUILD)/a028/9_14
+MACHINELEARNSET_DEPENDENCIES := data/generated/MachineMoveLearnsets.c
+MACHINELEARNSET_OBJS := $(patsubst data/generated/%.c,build/%.o,$(MACHINELEARNSET_DEPENDENCIES))
+MACHINELEARNSET_BIN := $(patsubst data/generated/%.c,build/%.bin,$(MACHINELEARNSET_DEPENDENCIES))
+
+$(MACHINELEARNSET_BIN): $(LEARNSETS_DATA)
```
zebben: these could be combined into a single call to speed up the process but it's not too slow as it is. could also make a `learnsets.mk` if preferred

scripts/build_learnsets.py
```diff
@@ -0,0 +1,250 @@
...
+    os.makedirs(os.path.dirname(output_path), exist_ok=True)
+
+    with open(output_path, "w") as f:
+        f.write("// DO NOT MODIFY THIS FILE! Autogenerated by build_learnsets.py\n\n")
+        f.write("#ifndef GENERATED_LEARNSET_CONSTANTS_H\n")
+        f.write("#define GENERATED_LEARNSET_CONSTANTS_H\n\n")
+        f.write(f"#define MAX_MACHINE_MOVES {max_num_machine_moves}\n")
+        f.write(f"#define MAX_LEVELUP_MOVES {max_num_levelup_moves}\n")
+        f.write(f"#define MAX_EGG_MOVES     {max_num_egg_moves}\n\n")
+        f.write(f"#define MACHINE_LEARNSETS_BITFIELD_COUNT (MAX_MACHINE_MOVES / 32)\n\n")
```
BluRosie: should be +1 to counter rounding down i believe

Makefile
```diff
@@ -267,6 +272,8 @@ all: $(TOOLS) $(OUTPUT) $(OVERLAY_OUTPUTS)
 	$(MAKE) move_narc
 	$(ARMIPS) armips/global.s
 	$(NARCHIVE) create $(FILESYS)/a/0/2/8 $(BUILD)/a028/ -nf
+	$(NARCHIVE) create $(FILESYS)/a/0/3/3 $(BUILD)/a033/ -nf
+	$(NARCHIVE) create $(FILESYS)/a/2/2/9 $(BUILD)/a229/ -nf
```
BluRosie: is there a reason these need to be done every time instead of just moving it like would otherwise be done?

scripts/build_learnsets.py
zebben: The num count is rounded up to nearest 32 so this should suffice I belive, but I think calculating it the way you're referring to is probably more clear. I'll update and verify

Makefile
zebben: Let me test this again. I had thought the narcs weren't actually copying to the rom properly without

scripts/build_learnsets.py
BluRosie: you can't store 132 bits in 4 u32's?

zebben: Oh I meant rounded up to nearest number *divisible* by 32 - so it's 128 with 100 TMs

BluRosie (approved):

PR #464, opened 2025-07-08
Anger shell

kumakuma1:
- add anger shell
- add subscript handling ability
- fix Weak Armor activation checks


data/battle_scripts/subscripts/subscript_0450_HANDLE_ANGER_SHELL.s
BluRosie, line 1: will merge and fix soon, but i'm pretty sure things tend to increase then decrease

PR #470, opened 2025-07-16
- add gem implementation

kumakuma1:
- adding subscript defines
- script to print message + remove item
- battlestruct flag
- check if move hits any battlers, then activate if any ist hit
- handle damage in CalcBaseDamage
- clear flag after Move

- TODO: symbiosis

Sidenote: any var *= y; statement does not work.

Rename any functions or variables if you like.


kumakuma1:
No longer need. See
https://github.com/BluRosie/hg-engine/pull/471/files

Issue #461, opened 2025-07-07
Beat Up uses the defenders living team members for number of hits


MichaelFacci:
Beat up is also using the incorrect text (says 7 times after 4 hits)

![Image](https://github.com/user-attachments/assets/d31b7e2a-f082-44f7-b0ac-940f4161130a)

PR #482, opened 2025-07-28
fix beat up

zebben:
Apologies, accidentally made the branch on upstream.

* fix hit count and display mismatches
* uses attacker's stat to calc damage 
  * https://bulbapedia.bulbagarden.net/wiki/Beat_Up_(move) 
  
Generation V onwards

Instead of using each party member's Attack stat, the user's Attack alone is used to calculate the damage of each strike. 
Because of this, Beat Up is now affected by boosts like from Swords Dance and Attack-boosting items like Choice Band. 
Furthermore, the base power per strike is no longer 10, but instead individually based on the Attack base stats of the party Pokémon:

BasePower=BaseAttackPartyMember10+5

https://github.com/user-attachments/assets/7b3b99ed-e3a7-4e9f-b555-dbf497466228

Addresses #461 

Closing because protected branch makes edits annoying.

https://github.com/BluRosie/hg-engine/pull/483

### DIFF
--- a/src/battle/battle_script_commands.c
+++ b/src/battle/battle_script_commands.c
@@ -1917,71 +1917,40 @@ BOOL btl_scr_cmd_6f_fury_cutter_damage_calc(void *bw UNUSED, struct BattleStruct
  */
 BOOL btl_scr_cmd_7c_beat_up_damage_calc(void *bw, struct BattleStruct *sp)
 {
-    int species, form, number_of_hits;
-    s32 newBaseDamage;
     struct PartyPokemon *mon;
 
     IncrementBattleScriptPtr(sp, 1);
 
     int partyCount = Battle_GetClientPartySize(bw, sp->attack_client);
 
+    // count the valid mons up front
     if (sp->multiHitCountTemp == 0) {
-
-        sp->multiHitCountTemp = 2;
+        sp->multiHitCount = 2;
         sp->loop_hit_check = 0xFD;
         sp->beat_up_count = 0;
-        mon = Battle_GetClientPartyMon(bw, sp->attack_client, sp->beat_up_count);
+        sp->damage_power = 5 + (sp->battlemon[sp->attack_client].attack / 10);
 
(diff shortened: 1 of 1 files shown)


src/battle/battle_script_commands.c
```diff
@@ -1917,71 +1917,40 @@ BOOL btl_scr_cmd_6f_fury_cutter_damage_calc(void *bw UNUSED, struct BattleStruct
...
         sp->loop_hit_check = 0xFD;
         sp->beat_up_count = 0;
-        mon = Battle_GetClientPartyMon(bw, sp->attack_client, sp->beat_up_count);
-
-        while(sp->beat_up_count != sp->sel_mons_no[sp->attack_client] &&
-                (GetMonData(mon, MON_DATA_HP, 0) == 0 ||
-                GetMonData(mon, MON_DATA_SPECIES_OR_EGG, 0) == 0||
-                GetMonData(mon, MON_DATA_SPECIES_OR_EGG, 0) == 494 ||
-                GetMonData(mon, MON_DATA_STATUS, 0) != 0))
+        sp->damage_power = 5 + (sp->battlemon[sp->attack_client].attack / 10);
```
Aeliko, line 1931: Shouldn't the base power be based on the party members' attack?

Also can we move the base power logic [here](https://github.com/BluRosie/hg-engine/blob/main/src%2Findividual%2FCalcBaseDamage.c#L575)?

zebben: included the excerpt from bulbapedia in the description but it's now based on the user's attack as of Gen V

I'll look at moving it there!

Aeliko: The attack stat of the attacker is used for damage calculation (need to turn on debug to see if the correct attack stat is used), but the move base power (`damage_power`) is determined by party members

zebben: ohhh gotcha. then yeah guess base damage needs to still like in this function I think

zebben: https://github.com/BluRosie/hg-engine/pull/483

PR #490, opened 2025-08-02
Retaliate

kumakuma1:
- bitmask to check if attacker slot has the fainted teammate
- hook faint animation function. All checkFainted..., etc lead to this function
- consolidate flags at end turn
- retaliate is boosted if a mon fainted the turn prior
- this works for singles and duoubles
- retaliate is not boosted if the partner faints in the same turn
- take party of fainted mon into account: works for 1on1, 1on2, 2on1 and 2on2, so that the fainted teammate only boosts retaliate in his party


include/battle.h
```diff
@@ -3332,6 +3337,11 @@ typedef enum Terrain {
 #define BATTLER_ENEMY2  3
 #define BATTLER_MAX     4
 
+// For setting a Bitmask to flag trainer position on enemy/player side
+#define TRAINER_1     1 //0b01
+#define TRAINER_2     2 //0b10
+#define TRAINER_BOTH  3 //0b11
```
Aeliko:
```suggestion
#define TRAINER_BOTH  (TRAINER_1 & TRAINER_2)
```

Aeliko (changes requested):

Aeliko (approved):

PR #491, opened 2025-08-04
fix First_Impression

kumakuma1:
- move failure is handled BeforeMove
image


Aeliko:
Can you add a new effect like how Aerial Ace is done so that the AI could recognize it in the future?

PR #496, opened 2025-08-05
Add Steam Engine

Pixelstyx:
Adds a new section to btl_scr_cmd_33_statbuffchange.c to account for raising a stat by 6 stages. Adds 3 entries to moves.c and battle_script_constants.h to account for the way the script IDs are shifted over. Adds a modified version of the code used for Rattled to MoveHitDefenderAbilityCheck.c.


include/constants/battle_script_constants.h
```diff
@@ -662,6 +662,9 @@
 #define ADD_STATUS_EFF_MORTAL_SPIN                     (189)
 #define ADD_STATUS_EFF_TIDY_UP                         (190)
 #define ADD_STATUS_EFF_BURN_AND_DRAIN_HEALTH           (191)
+#define ADD_STATUS_EFF_BOOST_STATS_ATTACK_UP_6         (192)
+#define ADD_STATUS_EFF_BOOST_STATS_DEFENSE_UP_6        (193)
+#define ADD_STATUS_EFF_BOOST_STATS_SPEED_UP_6          (194)
```
Aeliko: What are these used for?

Aeliko (approved):

include/constants/battle_script_constants.h
BluRosie: add these to mirrored asm constants at https://github.com/Pixelstyx/hg-engine/blob/main/asm/include/battle_constants.inc#L374

Pixelstyx: > What are these used for?

@Aeliko They're used to space out the script IDs so that they match the implementation of other stat stage changes in btl_scr_cmd_33_statbuffchange.c. Each stat has its own index number and the script gets them through the difference between the relevant attack script ID and the relevant stat script ID (in this case the difference is 2, which is the index for speed)

While I don't believe there are any other vanilla effects that alter stats by exactly six stages, this leaves the floor open to implement custom effects without much of a headache as opposed to just implementing speed up 6. It's a bit of a half-measure, since I only use the necessary spacer (defense up 6) and base (attack up 6), but I was wary of implementing more than were necessary. 

Since I need to make changes anyways, I'll add a comment below those lines to indicate that anyone wishing to add more +6 effects should move their IDs to the end of the list.

Pixelstyx: > add these to mirrored asm constants at https://github.com/Pixelstyx/hg-engine/blob/main/asm/include/battle_constants.inc#L374

@BluRosie Done! Totally missed that I needed to do that.

BluRosie: all good!  obviously the names aren't inherently indicative they should be correlated
might go around soon and fix that actually

BluRosie (approved):

Pixelstyx:
> would you mind merging in incinerate? if you don't mind otherwise i can manage it nbd

@BluRosie All conflicts with Incinerate have now been accounted for :)

PR #499, opened 2025-08-09
Fix Flower Veil & Protean (+Extras)

Pixelstyx:
BattleController_BeforeMove.c now actually checks for the target or allies to have the Grass type. btl_scr_cmd_33_statbuffchange now also utilizes HasType(), which notably accounts for type3 and active Tera type. This change was also applied to the catch formula for Net Balls, future-proofing wild pokemon Terastallization while keeping the door open for custom type3 setter moves.

Alters HasType() in other_battle_calculators.c to correctly ignore base types if the client is Terastallized for the sake of certain moves & abilities such as Prankster.

Protean and Libero now correctly have no effect if the user uses a typeless move or is Terastallized, and more importantly actually function if the move that would trigger them has no base power.

### DIFF
--- a/src/battle/other_battle_calculators.c
+++ b/src/battle/other_battle_calculators.c
@@ -3134,10 +3134,11 @@ int LONG_CALL GetClientActionPriority(struct BattleSystem *bsys UNUSED, struct B
 BOOL LONG_CALL HasType(struct BattleStruct *ctx, int battlerId, int type) {
     GF_ASSERT(TYPE_NORMAL < type && type < TYPE_STELLAR);
     struct BattlePokemon *client = &ctx->battlemon[battlerId];
-    return (client->type1 == type
+    return ((!(client->is_currently_terastallized) // Only check the client's base types if they are not terastallized. 
+         && (client->type1 == type
          || client->type2 == type
-         || client->type3 == type
-         || (client->is_currently_terastallized ? client->tera_type == type : FALSE));
+         || client->type3 == type))
+         || (client->is_currently_terastallized && client->tera_type == type));
 }
 
 
--- a/src/individual/btl_scr_cmd_33_statbuffchange.c
+++ b/src/individual/btl_scr_cmd_33_statbuffchange.c
@@ -251,9 +251,10 @@ BOOL btl_scr_cmd_33_statbuffchange(void *bw, struct BattleStruct *sp)
                 }
                 else if ((MoldBreakerAbilityCheck(sp, sp->attack_client, sp->state_client, ABILITY_FLOWER_VEIL) == TRUE
                        || MoldBreakerAbilityCheck(sp, sp->attack_client, BATTLER_ALLY(sp->state_client), ABILITY_FLOWER_VEIL) == TRUE) // any enemy has flower veil (accounting for mold breaker, otherwise would just CheckSideAbility)
-                      && (sp->battlemon[sp->state_client].type1 == TYPE_GRASS || sp->battlemon[sp->state_client].type2 == TYPE_GRASS)) // and target has grass type
+                       && HasType(sp, sp->state_client, TYPE_GRASS)) // and target has grass type
                 {
-                    // specifically for flower veil, we know that one of the Pokémon have flower veil.  we need to change the client that it prints the ability of to the flower veil client
+                    // The Pokémon with Flower Veil as an ability may not be the target of the move that triggers it. 
+                    // We need to make sure it prints the correct name and ability regardless by keeping track of the client with Flower Veil. 
                     u32 flower_veil_client;
 
                     flower_veil_client = (GetBattlerAbility(sp, sp->state_client) == ABILITY_FLOWER_VEIL) ? sp->state_client : BATTLER_ALLY(sp->state_client);
--- a/include/battle.h
+++ b/include/battle.h
@@ -1614,8 +1614,12 @@ enum
     BATTLE_MON_DATA_SLOW_START_COUNTER = 89,
 };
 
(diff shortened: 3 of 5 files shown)


Aeliko (approved):

Aeliko:
Where is the change for Soak?

Aeliko:
Note, if a client is Terastalized, it cannot be caught.

Pixelstyx:
@Aeliko 
> Where is the change for Soak?

I was typing without paying attention, sorry. This should have said 'Net Ball', which in my mind was related to Soak-- but since Soak changes type1 it was compatible regardless.

> Note, if a client is Terastalized, it cannot be caught.

This is what I get for not playing SV. Well, the change still allows for custom type3s at least!

data/battle_scripts/subscripts/subscript_0453_ADD_THIRD_TYPE_GRASS.s
```diff
@@ -0,0 +1,12 @@
+.include "asm/include/battle_commands.inc"
+
+.data
+
+_000:
+    Call BATTLE_SUBSCRIPT_ATTACK_MESSAGE_AND_ANIMATION
+    AddThirdType TYPE_GRASS
+    // Grass type was added to {0}!
+    PrintMessage 1579, TAG_NICKNAME, BATTLER_CATEGORY_DEFENDER // Sadly, there is no TAG_NICKNAME_MOVETYPE.
```
BluRosie: if you would like, i believe you could add it.  although not too many applicable situations so it may not be worth it

src/individual/btl_scr_cmd_33_statbuffchange.c
```diff
@@ -251,9 +251,10 @@ BOOL btl_scr_cmd_33_statbuffchange(void *bw, struct BattleStruct *sp)
                 }
                 else if ((MoldBreakerAbilityCheck(sp, sp->attack_client, sp->state_client, ABILITY_FLOWER_VEIL) == TRUE
                        || MoldBreakerAbilityCheck(sp, sp->attack_client, BATTLER_ALLY(sp->state_client), ABILITY_FLOWER_VEIL) == TRUE) // any enemy has flower veil (accounting for mold breaker, otherwise would just CheckSideAbility)
-                      && (sp->battlemon[sp->state_client].type1 == TYPE_GRASS || sp->battlemon[sp->state_client].type2 == TYPE_GRASS)) // and target has grass type
+                       && HasType(sp, sp->state_client, TYPE_GRASS)) // and target has grass type
                 {
-                    // specifically for flower veil, we know that one of the Pokémon have flower veil.  we need to change the client that it prints the ability of to the flower veil client
+                    // The Pokémon with Flower Veil as an ability may not be the target of the move that triggers it.
```
BluRosie: if this is the case, will need to change how the `hasFlowerVeil` check is done elsewhere to not check defender for flower veil

src/battle/other_battle_calculators.c
```diff
@@ -1629,6 +1654,22 @@ int LONG_CALL ServerDoTypeCalcMod(void *bw UNUSED, struct BattleStruct *sp, int
                     }
                 }
             }
+            if ((TypeEffectivenessTable[i][1] == defender_type_3) && (defender_type_3 != defender_type_1) && (defender_type_3 != defender_type_2))
+            {
+                if (ShouldUseNormalTypeEffCalc(sp, attack_client, defence_client, i) == TRUE
+                && !(!CheckSideAbility(bw, sp, CHECK_ABILITY_ALL_HP, 0, ABILITY_CLOUD_NINE)
+                    && !CheckSideAbility(bw, sp, CHECK_ABILITY_ALL_HP, 0, ABILITY_AIR_LOCK)
+                    && sp->field_condition & WEATHER_STRONG_WINDS
+                    && (TypeEffectivenessTable[i][2] == 20)
```
BluRosie: replace with constant

BluRosie (changes requested):

src/battle/other_battle_calculators.c
Pixelstyx: @BluRosie Replace what in particular?

src/individual/btl_scr_cmd_33_statbuffchange.c
Pixelstyx: @BluRosie I'll dig into that and ensure hasFlowerVeil is set up correctly.

data/battle_scripts/subscripts/subscript_0453_ADD_THIRD_TYPE_GRASS.s
Pixelstyx: @BluRosie Honestly, I can't think of a single circumstance where it would be used. While I could implement it, it was technically a shortcut to cut down on implementation for custom moves. The true internal effect of adding a third type in modern gens is probably either a param (doubtful) or move-specific. We'll stick with the latter (this) and I'll cut my bellyaching from the comments.

src/individual/btl_scr_cmd_33_statbuffchange.c
Pixelstyx: So all of this code is wrong, flower veil blocking any stat drop or status effect outside of a secondary effect will play the line "{DEFENDER} surrounded themselves with a veil of petals!". Secondary effect (tested on Electroweb) plays nothing at all.

Pixelstyx: I wonder if the original commenter wanted to make a battle fly-in and this code was forgotten...

src/battle/other_battle_calculators.c
Pixelstyx: Ah I understand.

PR #508, opened 2025-08-16
simplify learnsets setup

zebben:
- use `data/learnsets/learnsets.json` for the source of mon learnsets instead of the complicated inheritance stuff
- script still supports inheritance setups to tailor learnsets based on user input
  - added a readme to describe this
- ensure the py script runs before C compilation so that `learnsets.h` is up-to-date before compilation
- remove generated files from git. may add back if discussion leans that way


data/learnsets/base/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+This woud use pure HGSS learnsets meaning any newer mons would not have any learnable moves. Normally it will be better to simply take the hgss json file and copy that into `data/learnsets/learnsets.json`
+
+```bash
+python3 scripts/build_learnsets.py --generate data/learnsets/learnsets.json --cutoff hgss
+```
+
+#### National Dex (Repo Default)
+
+Takes the latest available maingame learnsets for each mon. For example, Raticate would take it's BDSP level up learnset because it doesn't have learnsets for later gens.
```
Aeliko:
```suggestion
Takes the latest available maingame learnsets for each mon. For example, Raticate would take its BDSP level up learnset because it doesn't have learnsets for later gens. 
```

PR #516, opened 2025-08-21
Flying Press Draft, WIP

kumakuma1:
- use defines in typeEffectivnessTable
- you can ignore the changes in the Table if you like the previous version better
- simplify type effectiveness check via multiplication
- add quadruple effectiveness defines for stuff later
- TODO: type3 effectiveness
Unbenannt


kumakuma1:
> Does this affect the "switch-in AI"?

Not yet ;-)

PR #522, opened 2025-08-27
Update logic to consider Mold Breaker for some abilities

Aeliko:


src/battle/ability.c
```diff
@@ -417,6 +417,104 @@ BOOL LONG_CALL MoveHitDefenderAbilityCheck(void *bw, struct BattleStruct *sp, in
     return ret;
 }
 
+/// @brief Check if ability can be disabled by Mold Breaker
+/// @param ability
+/// @return `TRUE` or `FALSE`
+BOOL AbilityBreakable(int ability) {
```
BluRosie: interested in seeing how big this function is.  may just make a bitfield for this or something

.clang-format
BluRosie, line 1: OwO what's this?

src/battle/other_battle_calculators.c
```diff
@@ -2351,20 +2354,20 @@ BOOL LONG_CALL BattleSystem_CheckMoveEffect(void *bw, struct BattleStruct *sp, i
...
+            return TRUE;
         }
     }
 
     if (sp->battlemon[battlerIdTarget].effect_of_moves & MOVE_EFFECT_FLAG_MINIMIZED
         && !sp->battlemon[battlerIdTarget].is_currently_dynamaxed
         && IsMoveInMinimizeVulnerabilityMovesList(move)) {
             sp->waza_status_flag &= ~MOVE_STATUS_FLAG_MISS;
-            return FALSE;
+            return TRUE;
```
BluRosie, line 2371: what do the TRUE/FALSE signal here?  is it being handled differently above?

BluRosie: i remember these being changed to FALSE and also being confused then

src/battle/battle_script_commands.c
```diff
@@ -2004,20 +2003,26 @@ const u16 sLowKickWeightToPower[][2] =
...
+            weight *= 2;
+        } else if (GetBattlerAbility(sp, client) == ABILITY_LIGHT_METAL) {
+            weightModifier *= 2;
+        }
+    } else {
+        if (MoldBreakerAbilityCheck(sp, attack_client, client, ABILITY_HEAVY_METAL)) {
+            weight *= 2;
+        } else if (MoldBreakerAbilityCheck(sp, attack_client, client, ABILITY_LIGHT_METAL)) {
+            weightModifier *= 2;
+        }
```
BluRosie, line 2025: add a check to make sure that attack_client != client (unless if an attacker can cancel its own weight for e.g. heavy slam)
EDIT:  clients can not have two different abilities, so if it had mold breaker it wouldn't have heavy metal.  leaving this here because it's funny

src/individual/CalcBaseDamage.c
```diff
@@ -996,7 +996,7 @@ int UNUSED CalcBaseDamage(void *bw, struct BattleStruct *sp, int moveno, u32 sid
 
             // handle Punk Rock
             // TODO: confirm location
-            if (AttackingMon.ability == ABILITY_PUNK_ROCK && IsMoveSoundBased(sp->current_move_index)) {
+            if (MoldBreakerAbilityCheck(sp, attacker, attacker, ABILITY_PUNK_ROCK) && IsMoveSoundBased(sp->current_move_index)) {
```
BluRosie: is it fair to not use MoldBreakerAbilityCheck for the attacker?

src/individual/MoveHitDefenderAbilityCheck.c
BluRosie, line 1: this diff sucks.  is it a clang-format diff?  would prefer to keep most of those changes to a separate pr or something so that i don't fr have to review it for changes.  i will review this when i am not doing so covertly

.clang-format
Aeliko: I took it from pokeplatinum

src/battle/other_battle_calculators.c
Aeliko: TRUE -> must hit so can skip accuracy check
FALSE -> consume RNG call to check accuracy

src/battle/battle_script_commands.c
Aeliko: The function now handles if attacker equals defender

Trying to make all ability checks use this function so that edge cases involving Mold Breaker would just work

src/individual/MoveHitDefenderAbilityCheck.c
Aeliko: No since we need to check for Mold Breaker for a lot of the cases I changed to if-else instead of switch-case

.clang-format
BluRosie: how do i get it to be used and enforced throughout the repository?  since we want to enforce a style guide

include/battle.h
```diff
@@ -1959,6 +1959,87 @@ enum {
 //     TRY_MOVE_END,
 // };
 
+struct PACKED sDamageCalc
+{
+    u16 species;
+    s16 hp;
+    u16 maxhp;
+    u16 dummy;
+    int item;
```
zebben: couple of ints here. u16 / u 16 / u8?

include/battle.h
```diff
@@ -1959,6 +1959,87 @@ enum {
...
+
+    u32 attack;
+    u32 defense;
+    u32 sp_attack;
+    u32 sp_defense;
+    s8 atkstate;
+    s8 defstate;
+    s8 spatkstate;
+    s8 spdefstate;
+    u8 positiveStatBoosts;
```
zebben, line 1994: mix of snake and camel case going on here. probably a good time to standardize. DamageCalcStruct too

src/battle/ability.c
```diff
@@ -419,6 +419,134 @@ BOOL LONG_CALL MoveHitDefenderAbilityCheck(void *bw, struct BattleStruct *sp, in
...
```
zebben: super nit especially because you mostly just moved this but maybe `AbilityIsIgnoredByMoldBreaker` or something

src/battle/battle_calc_damage.c
```diff
@@ -46,17 +46,103 @@ const u8 StatBoostModifiers[][2] = {
...
+    damageCalc.terrainOverlayNumberOfTurnsLeft = sp->terrainOverlay.numberOfTurnsLeft;
+
+    damageCalc.playerSideHasFaintedTeammateLastTurn = sp->playerSideHasFaintedTeammateLastTurn;
+    damageCalc.enemySideHasFaintedTeammateLastTurn = sp->enemySideHasFaintedTeammateLastTurn;
+
+    damageCalc.originalMoveType = sp->moveTbl[moveno].type;
+    damageCalc.moveEffect = sp->moveTbl[moveno].effect;
+    damageCalc.moveFlag = sp->moveTbl[moveno].flag;
+
+        for (u32 i = 0; i < damageCalc.maxBattlers; i++) {
```
zebben: untab this line

.clang-format
zebben: agreed this would be nice

I think we can do a PR action like (untested)

```yaml
name: Clang Format Check
on: [pull_request]

jobs:
  clang-format:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: Install clang-format
        run: sudo apt-get install -y clang-format
      - name: Check formatting
        run: |
          FILES=$(git ls-files ''src/*.c' 'include/*.h')
          clang-format --dry-run --Werror $FILES
```

I believe many editors will honor it if it exists in the repo automatically. we could add a pre commit hook too

```
cat > .git/hooks/pre-commit <<'EOF'
#!/bin/sh
files=$(git diff --cached --name-only --diff-filter=ACM | grep -E '\.(c|h)$')
[ -z "$files" ] && exit 0

clang-format -i $files
git add $files
EOF
chmod +x .git/hooks/pre-commit
```

src/individual/MoveHitDefenderAbilityCheck.c
```diff
@@ -34,603 +34,484 @@ BOOL MoveHitDefenderAbilityCheckInternal(void *bw, struct BattleStruct *sp, int
...
-             && sp->moveTbl[sp->current_move_index].power != 0
-             && BattleRand(bw) % 10 < 3)
-            {
-                sp->waza_work = sp->current_move_index;
-                sp->battlemon[sp->attack_client].moveeffect.disabledMove = sp->waza_work;
-                sp->battlemon[sp->attack_client].moveeffect.disabledTurns = 4; // cursed body disables for 4 turns every time
+        }
+    } else if (MoldBreakerAbilityCheck(sp, sp->attack_client, sp->defence_client, ABILITY_JUSTIFIED)) {
+        if ((sp->battlemon[sp->defence_client].hp)
+            && (sp->battlemon[sp->defence_client].states[STAT_SPEED] < 12)
```
zebben:
```suggestion
            && (sp->battlemon[sp->defence_client].states[STAT_ATTACK] < 12)
```

src/individual/MoveHitDefenderAbilityCheck.c
```diff
@@ -34,603 +34,484 @@ BOOL MoveHitDefenderAbilityCheckInternal(void *bw, struct BattleStruct *sp, int
...
-            }
-            break;
-        case ABILITY_SAND_SPIT:
+                ((sp->battlemon[sp->defence_client].hp - (sp->oneSelfFlag[sp->defence_client].physical_damage)) > (s32)sp->battlemon[sp->defence_client].maxhp / 2) || ((sp->battlemon[sp->defence_client].hp - (sp->oneSelfFlag[sp->defence_client].special_damage)) > (s32)sp->battlemon[sp->defence_client].maxhp / 2))) {
+            sp->addeffect_type = ADD_EFFECT_ABILITY;
+            sp->state_client = sp->defence_client;
+            sp->battlerIdTemp = sp->defence_client;
+            seq_no[0] = SUB_SEQ_HANDLE_ANGER_SHELL;
+            ret = TRUE;
+        } else if (MoldBreakerAbilityCheck(sp, sp->attack_client, sp->defence_client, ABILITY_SAND_SPIT)) {
```
zebben: I believe this is being elsed within the anger shell conditional, needs to be brought back out 1 layer

src/battle/battle_calc_damage.c
```diff
@@ -46,17 +46,103 @@ const u8 StatBoostModifiers[][2] = {
...
+        client.level = BattlePokemonParamGet(sp, i, BATTLE_MON_DATA_LEVEL, NULL);
+        client.species = BattlePokemonParamGet(sp, i, BATTLE_MON_DATA_SPECIES, NULL);
+        client.hp = BattlePokemonParamGet(sp, i, BATTLE_MON_DATA_HP, NULL);
+        client.maxhp = BattlePokemonParamGet(sp, i, BATTLE_MON_DATA_MAX_HP, NULL);
+        client.condition = BattlePokemonParamGet(sp, i, BATTLE_MON_DATA_MAX_CONDITION, NULL);
+        client.condition2 = sp->battlemon[i].condition2;
+        client.ability = GetBattlerAbility(sp, i);
+        client.sex = BattlePokemonParamGet(sp, i, BATTLE_MON_DATA_SEX, NULL);
+        client.speed = sp->effectiveSpeed[i];
+        client.weight = GetPokemonWeight(bw, sp, i, i);
```
zebben: should it be `bw, sp, attacker, i`?

src/battle/battle_calc_damage.c
```diff
@@ -46,17 +46,103 @@ const u8 StatBoostModifiers[][2] = {
 int CalcBaseDamage(void *bw, struct BattleStruct *sp, int moveno, u32 side_cond,
```
zebben, line 46: unused params:
- `pow`
- `side_cond`
- `field_cond`

is this right? I see the struct has the latter 2 defined but not passing it from the param

.clang-format
BluRosie: @zebben could i get you to work on that in a pr or similar?  i haven't ever done anything with github actions or anything like that

BluRosie:
fix weight param (and merge conflict) and this will be set for merge 👍

PR #529, opened 2025-08-31
Trace, Sticky Web, no target crash, Fix Dig into Magnet Rise/Levitate bug

### LINKED ISSUE #425: explosion -> mon survives crashes the game
https://discord.com/channels/446824489045721090/1201213967389966378/1373383777023295649
my understanding is that both...
- no target left crashes the game
- targeting the fainted mon fails to redirect and crashes the game

### LINKED ISSUE #484: Dig creates softlock when used on an air balloon mon
Image

If you use an invulnerable move like dig and it fails, you don't become vulnerable again, and are stuck in the dig second turn every turn. This only happens if dig fails (like on an air balloon here), but not if it connects.

This creates a soft lock since the enemy spams moves into the invulnerable pokemon and you cant select another move since you are technically still on turn 2 of dig. It's probably a flag somewhere that just isnt being reset.

Also note this bug does not happen against flying types like skarmory, so I am guessing it's an air balloon bug?

Image

kumakuma1:
Ability gets overridden, so removing sp->battlemon[client_no].ability_activated_flag = 1; will trigger the new ability in SwitchInAbilityCheck, #507

Add to sticky web checks, #524

fix #425 on melonds

fix #484


data/battle_scripts/subscripts/subscript_0099_HAZARDS_CHECK.s
BluRosie, line 1: did we have a way now to check type 3 in battle scripts?

kumakuma1: Would probably need a battle script command?!

kumakuma1:
> just to make sure--are these tested?

Made a video here
https://discord.com/channels/446824489045721090/1201217040824942602/1411758305453412535

data/battle_scripts/subscripts/subscript_0099_HAZARDS_CHECK.s
BluRosie: sounds good.  i suppose we'll get there when it's really relevant

PR #532, opened 2025-09-04
Implement Aurora Veil & Fix Lure, Safari and Sport Ball Rates

Pixelstyx:
Still needs hooks to account for TryBreakScreens and possibly GetHeldItemModifier.


src/individual/CalculateBallShakes.c
Aeliko, line 1: Add a default off config to revert back to Gen 4 rates

Aeliko (changes requested):

src/individual/CalculateBallShakes.c
Pixelstyx: No.

PR #536, opened 2025-09-09
basic functionality for rotom catalog

zebben:
cycles through the formes rather than showing a picker since this was much easier. can discard this if we don't want it

https://github.com/user-attachments/assets/bf9b2a0b-3f07-46d9-8bb0-f3eb9861d6f7


include/pokemon.h
```diff
@@ -1889,4 +1889,6 @@ u8 LONG_CALL GetMoveMaxPP(u16 moveId, u8 ppUps);
 
 void LONG_CALL ApplyMonMoodModifier(struct PartyPokemon *mon, int modifierId);
 
+BOOL Mon_UpdateRotomForm(struct PartyPokemon *mon, int form, int defaultSlot);
```
BluRosie: would need LONG_CALL

zebben: added and fixed conflicts

PR #557, opened 2025-10-09
Machine Move QOL

zebben:
- script/make command to update item descriptions and sprites based on the selected moves. run `make update_machine_moves`
- added fairy tm icon (thanks brioche for the base image. I just had to convert it to the right format)
- more closely match the TM usage in S/V by removing 2 redundant yes/no confirmations
- more closely match the TM usage text in S/V

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -482,9 +482,6 @@ dumprom: $(VENV_ACTIVATE)
 	$(MAKE) clean
 	chmod +x $(DUMP_SCRIPT_LOCATION)/*.sh
 
-# dump mondata, encounters, evos, moves
-	$(PYTHON) tools/source/dumptools/dump_narcs.py $(ROMNAME)
-
 	./$(DUMP_SCRIPT_LOCATION)/dumprom.sh
 	mkdir -p $(BUILD) $(BUILD_NARC) $(BUILD)/a028/
 # dump human overworlds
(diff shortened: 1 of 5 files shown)


data/text/010.txt
```diff
@@ -57,9 +57,9 @@ Is it OK to throw away\n{STRVAR_1 52, 1, 0}?
                                          
 You can’t dismount your Bike here.
 Your Coins: {STRVAR_1 54, 0, 0}
-Which Pokémon will you use it on?\n{WAIT 3}{WAIT 4}
-Which Pokémon will you use it on?\n{WAIT 3}{WAIT 4}
-Please choose a move that will be\nreplaced with {STRVAR_1 6, 0, 0}.
+You booted up the TM.{WAIT 3}{WAIT 4}\n
```
zebben, line 60: Restoring this back to what it was before because they are now skipped anyway.

data/codetables.mk
```diff
@@ -78,7 +78,7 @@ TUTORLEARNSET_DEPENDENCIES := $(LEARNSET_OUTPUT_DIR)/TutorMoveLearnsets.c
 LEVELUPLEARNSET_DEPENDENCIES := $(LEARNSET_OUTPUT_DIR)/LevelupLearnsets.c
 EGGLEARNSET_DEPENDENCIES := $(LEARNSET_OUTPUT_DIR)/EggLearnsets.c
 
-$(LEARNSETS_HEADER): $(LEARNSETS_INPUT) $(VENV_ACTIVATE)
+$(LEARNSETS_HEADER): $(LEARNSETS_INPUT) $(VENV_ACTIVATE) src/item.c
```
zebben, line 81: so if a user updates `sMachineMoves` it will update learnsets too without also having to touch the json file

PR #569, opened 2025-11-02
Fix go sync

zebben:
- fix deps for go sync
- only run on BluRosie/hg-engine
- consolidate the item max constants
- placeholder text to the archives for items

example PR:
https://github.com/zebben/hg-engine/pull/3

### DIFF
--- a/src/item.c
+++ b/src/item.c
@@ -427,10 +427,6 @@ u16 GetItemIndex(u16 item, u16 type)
             break;
         }
         ret = item;
-        //if (item > MAX_TOTAL_ITEM_NUM)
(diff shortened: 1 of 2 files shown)


src/item.c
```diff
@@ -484,7 +483,7 @@ void *GetItemArcData(u16 item, u16 type, u32 heap_id)
 {
     int dataid, picid,palid;
 
-    //if (item > ITEM_DATA_MAX)
+    //if (item > MAX_TOTAL_ITEM_NUM)
```
BluRosie: would prefer to delete these comments than to get rid of what they were originally representing by changing this (that being the original implementation of the expanding where only expanded items were mapped to their data files by index as opposed to current implementation where all items are mapped to their data files by index)

zebben: should I get rid of these and the `GetItemIndex` comments as well?

BluRosie (approved):

PR #572, opened 2025-11-07
The Great Battle Script Cleanup

Pixelstyx:
Every single subscript has had its formatting changed, consts and comments added to better describe what's happening.

Most fixes that were included in this PR were merged elsewhere.


data/battle_scripts/effects/effect_script_0010_ATK_UP.s
mozzydippers, line 1: now only called by sharpen since howl got moved into its own subscript bc of modernization & such

data/battle_scripts/effects/effect_script_0101_LEAVE_WITH_1_HP.s
```diff
@@ -2,7 +2,8 @@
 
 .data
 
-_000:
+// Called by False Swipe, for some reason.
```
mozzydippers: should also be called by hold back if it isnt already

data/battle_scripts/effects/effect_script_0103_PRIORITY_1.s
```diff
@@ -2,7 +2,9 @@
 
 .data
 
-_000:
+// Called by Accelerock, Aqua Jet, Bullet Punch, Grassy Glide, Ice Shard, Jet Punch, Mach Punch, Quick Attack, Shadow Sneak and Vacuum Wave.
+// ...for some reason.
```
mozzydippers: pretty sure priority 1 is an eff less so bc the eff actually does anything but just to signal to ai ? atleast i presume

data/battle_scripts/effects/effect_script_0111_PROTECT.s
```diff
@@ -2,6 +2,7 @@
 
 .data
 
-_000:
+// Called by Protect and Detect.
```
mozzydippers, line 5: should probably note that stuff like king shield & baneful bunker also call the protect subscript but the lowering atk / poison is handled elsewhere

data/battle_scripts/effects/effect_script_0140_RAISE_ALL_STATS_HIT.s
```diff
@@ -2,7 +2,8 @@
 
 .data
 
-_000:
+// Called by Ancient Power, Ominous Wind and Silver Wind.
```
mozzydippers, line 5: clangorous soulblaze presuming z moves dont get their own effects

data/battle_scripts/effects/effect_script_0178_COPY_ABILITY.s
mozzydippers, line 1: most of the checks in role play skill swap etc can be removed in cleanup given theyre present in beforemove

data/battle_scripts/effects/effect_script_0198_RECOIL_THIRD.s
```diff
@@ -2,11 +2,12 @@
 
 .data
 
-_000:
-    CheckAbility CHECK_OPCODE_NOT_HAVE, BATTLER_CATEGORY_ATTACKER, ABILITY_RECKLESS, _008
+// Called by Brave Bird, Double-Edge, Flare Blitz, Volt Tackle, Wave Crash and Wood Hammer.
```
mozzydippers: flare blitz is subscript 253 & volt tackle is 262

data/battle_scripts/effects/effect_script_0010_ATK_UP.s
Pixelstyx: Fixed, forgot to go back for that one.

data/battle_scripts/effects/effect_script_0103_PRIORITY_1.s
Pixelstyx: This is what I am gathering from reading some other spots in the code-- it seems like eff IDs are used elsewhere in the code to check whether things like Substitute should work. Until we know where, though, I don't want to pretend we do.

data/battle_scripts/effects/effect_script_0111_PROTECT.s
Pixelstyx: For indexing purposes I try not to namedrop moves not called by an effect script. If we're being thorough and the others aren't handled here, it should be implicit. Also, the user should be able to look up King's Shield or Baneful Bunker and figure out where it is (or isn't)

data/battle_scripts/effects/effect_script_0140_RAISE_ALL_STATS_HIT.s
Pixelstyx: It might end up being its own effscript since it needs to do other things to 'use up' the z-crystal, but we could comment it the way I did with a few G-Max Moves to say we're not sure yet.

data/battle_scripts/effects/effect_script_0178_COPY_ABILITY.s
Pixelstyx: I was advised by Blu not to touch 99% of fail checks to not even further complicate this pass. We'll handle them after for certain.

data/battle_scripts/effects/effect_script_0198_RECOIL_THIRD.s
Pixelstyx: Oops. Those will be fixed in the upcoming commit. When I first got here it hadn't quite sunk in how eff scripts work and I hadn't returned to it.

data/battle_scripts/effects/effect_script_0111_PROTECT.s
Pixelstyx: Unless you meant that they call the effect script and the secondary effects are handled elsewhere as they don't occur instantly?

data/battle_scripts/effects/effect_script_0101_LEAVE_WITH_1_HP.s
Pixelstyx: Gonna be real with you, I forgot that move existed and didn't bother to check for variants of false swipe.

data/battle_scripts/effects/effect_script_0111_PROTECT.s
Pixelstyx: Huh, they do work like that. I will fix that up.

PR #576, opened 2025-11-14
Battle simulation

zebben:
### Running it
1. uncomment flag in `debug.h`
2. set `ACTIVE_TEST_SCENARIO` in `test_battle.c`
3. build then load into any trainer battle

### TODOs
* better integration with an emulator to automate testing


Aeliko:
I'm thinking whether to combine aiScript and aiTarget into 1 struct

Aeliko (commented):
I forgot to use this instead

zebben:
Marked as ready for review. I think the last major piece is gonna be emulator integration to further automate stuff but this should already be a nice help for testing I think

BluRosie:
actually going to merge now so that i don't have to solve 30 million conflicts.  but we need to figure out how to get the tests set up

PR #582, opened 2025-11-26
add rock smash item table repoints

kumakuma1:


Aeliko (commented):
Can the number of entries be increased? If not better add some comments mentioning that

BluRosie:
going to add a table that expands the necessary part of `Task_GetRockSmashItem` so that this can be neatly expanded 👍

PR #583, opened 2025-11-26
Battle BG Expansion + Map Weather Terrain at battle start

zebben:
- New battle bgs are set before the battle actually starts so there isn't as much visible flash on bg change
- Allow for map headers for thunderstorm to create permanent rain + electric terrain at battle start
- Allow for map headers for mist to create permanent misty terrain at battle start
- Should allow for people to add custom battle bg logic
- Tested with Grassy Terrain being set up via move after start with Electric and it expires fine without causing issues as far as I can tell

Graphics glitch for the trainer backsprite. I couldn't figure out how to actually solve this one unfortunately.

https://github.com/user-attachments/assets/55bcfeaf-8fb5-406a-bafd-9c4e79db3fa6

https://www.serebii.net/swordshield/weather.shtml
Sw/Sh:
- Thunderstorm | Rain Dance is in effectElectric Terrain is in effect
- Fog | Misty Terrain is in effect

### DIFF
--- a/include/battle.h
+++ b/include/battle.h
@@ -3973,7 +3973,7 @@ BOOL StrongWindsShouldWeaken(struct BattleSystem *bw, struct BattleStruct *sp, i
  * @brief Inject a custom callback function to allow
  * loading new battle bgs at the start of a battle
  */
-void LONG_CALL BattleBgExpansionLoader();
+void LONG_CALL BattleBgExpansionLoader(struct BattleSystem *bsys);
 
 /**
  * @brief Callback for loading custom battle backgrounds


src/battle/battle_start.c
```diff
@@ -37,6 +37,8 @@ struct BattleStruct *ServerInit(struct BattleSystem *bw) {
     ServerMoveAIInit(bw, sp);
     DumpMoveTableData(&sp->moveTbl[0]);
     sp->aiWorkTable.item = ItemDataTableLoad(HEAPID_BATTLE_HEAP);
+    bw->sp = sp;
```
zebben: @BluRosie do you think this will cause issues? I'm not sure why bw doesn't have sp yet at this point but the loader logic requires sp so we can set weather and terrain up

src/battle/battle_input.c
```diff
@@ -851,15 +852,105 @@ void LoadDifferentBattleBackground(struct BattleSystem *bw, u32 bg, u32 terrain)
...
+    switch (weather) {
+    case WEATHER_SYS_STORM:
+    case WEATHER_SYS_THUNDER:
+        terrainType = ELECTRIC_TERRAIN;
+        bsys->sp->field_condition &= ~FIELD_CONDITION_WEATHER;
+        bsys->sp->field_condition |= WEATHER_RAIN_PERMANENT;
+        break;
+    case WEATHER_SYS_MIST1:
+    case WEATHER_SYS_MIST2:
+        terrainType = MISTY_TERRAIN;
```
Aeliko, line 896: Add config for either Fog or Misty Terrain, not both at the same time

Aeliko (changes requested):
Remove debug prints and testing code

src/battle/battle_start.c
BluRosie: it shouldn't cause issues i wouldn't think

include/battle.h
```diff
@@ -3948,5 +3955,15 @@ int GetSanitisedType(int type);
 
 BOOL StrongWindsShouldWeaken(struct BattleSystem *bw, struct BattleStruct *sp, int typeTableEntryNo, int defender_type);
 
+/**
+ * @brief Inject a custom callback function to allow
+ * loading new battle bgs at the start of a battle
+ */
+void LONG_CALL BattleBgExpansionLoader();
```
Aeliko:
```suggestion
void LONG_CALL BattleBgExpansionLoader(struct BattleSystem *bsys);
```

PR #601, opened 2025-12-09
default new terrain weather configs to off

zebben:
- Defaults weather-terrain configs to off
- Removes the artifacts from the slide-in animation

https://github.com/user-attachments/assets/9759e45d-2a52-4b7e-bad2-8efa362186e1


Aeliko (approved):

BluRosie:
the 512x32 is just how it needed to happen to fill in the vram the platform should have occupied?

PR #602, opened 2025-12-10
Fix Rollout, Fury Cutter, Clear Smog

Aeliko:


armips/data/moves.s
```diff
@@ -14417,7 +14413,7 @@ movedata MOVE_BLAZING_TORQUE, "Blazing Torque"
     appeal 0x00
     contesttype CONTEST_COOL
     terminatedata
-    movedescription MOVE_BLAZING_TORQUE, "The user rams into the\ntarget with blazing\nenergy. This may also\nleave the target with\na burn."
+    movedescription MOVE_BLAZING_TORQUE, "---"
 
 movedata MOVE_WICKED_TORQUE, "Wicked Torque"
```
BluRosie, line 14418: what are we doing here gang

src/individual/CalcBaseDamage.c
```diff
@@ -359,13 +359,19 @@ int UNUSED CalcBaseDamageInternal(struct BattleSystem *bw, struct BattleStruct *
...
         }
         break;
     case MOVE_ROLLOUT:
     case MOVE_ICE_BALL:
-        // CalcRolloutPower decrements rolloutCount before running this loop, meaning it will range between 4-0.
+        // TODO: Handle Rollout storage, need to hook `BtlCmd_CalcRolloutPower`
+        // For now we use the Gen 4 implementation which is basically the same
+        // Edit: Rollout storage seems to be patched
+        // https://github.com/pret/pokeheartgold/blob/29282f7bb45946dee63475022a8d506092bc3748/src/battle/battle_command.c#L3391
         for (u32 n = 0; n < 4 - AttackingMon.rolloutCount; n++) {
```
BluRosie, line 375: should this be 4 or 5?  my apologies

BluRosie: should be n = 1; n < 5?  will fix on merge

armips/data/moves.s
Aeliko: There's no description in SV. This is the canonical description.

Aeliko: There's no description in SV. This is the canonical description.

src/individual/CalcBaseDamage.c
Aeliko: This should be correct it's the vanilla implementation. Nothing has changed from Gen 4 to 9. The counter should start at 0 instead of 1.

BluRosie (approved):

PR #611, opened 2025-12-15
swap apricorns back

zebben:
The dumped descriptions and everything are already pointing at the original vanilla order. So we either fix the bug of the descriptions and stuff being wrong or the item id being wrong. imo just go back to the original hgss order as it seems like a silly mistake on GF's end to swap them in the first place


Aeliko (commented):
This PR changes back to latest indices. However, people will complain when vanilla scripts, Apricorn Box, Apricorn trees behaviour "changes".

zebben:
I'll fix descriptions then reopen then.

PR #617, opened 2025-12-18
increase script buffer and fix battlerId for type3 check

zebben:


Aeliko (approved):

Pixelstyx:
we should really make sure this finds its way into item-dev as well since it is broken there too.

BluRosie (approved):

PR #620, opened 2025-12-18
Small fixes, remove obediance in test battles

kumakuma1:
- remove obediance check in tests
- small fixes for flame orb, dream eater


src/test_battle.c
BluRosie, line 1: how about we set up a folder for these to go into and then a file that includes all of them that src/test_battle.c actually includes?  might make it such that the test cases are read from a narc instead

zebben: > might make it such that the test cases are read from a narc instead

I'm looking at this right now and trying to get them to run sequentially

PR #621, opened 2025-12-19
load from narc and start tests one after the other

### LINKED ISSUE #609: we need to figure out how to add battle tests in emulator
we have the capability to create completely automated battles that start immediately without input.  we need to start defining tests for battle scenarios

zebben:


include/battle.h
```diff
@@ -4042,11 +4042,16 @@ void LONG_CALL BattleBgExpansionLoader(struct BattleSystem *bsys);
 void LONG_CALL BattleBackgroundCallback(void *unkPtr, UNUSED int unk2, UNUSED int unk3);
 
 #ifdef DEBUG_BATTLE_SCENARIOS
+BOOL LONG_CALL TestBattle_HasMoreTests();
```
zebben: Move these to the new header file to try to isolate further

requirements.txt
```diff
@@ -1,2 +1,4 @@
 ndspy==4.1.0
 pandas
+Pillow
+py-desmume
```
Aeliko, line 4: I just global installed, not sure if this works with the whole environment thingy.

scripts/run_tests.py
```diff
@@ -0,0 +1,28 @@
+from desmume.emulator import DeSmuME
+from desmume.emulator import DeSmuME_Memory
+
+SHOW_VIDEO_OUTPUT = False
+g_EmulatorCommunicationSendHoleAddress = 0x023df15c
+NUMBER_OF_TESTS_TO_RUN = 1
```
Aeliko: Need to improve this.

scripts/run_tests.py
```diff
@@ -0,0 +1,28 @@
...
+# Run the emulation as fast as possible until quit
+while True:
+    window.process_input()   # Controls are the default DeSmuME controls, see below.
+    emu.cycle()
+    if (SHOW_VIDEO_OUTPUT):
+        window.draw()
+    # TODO: Obviously make it not 1
+    # Kill the emulator after running 1 test case
+    if (emu_memory.unsigned[g_EmulatorCommunicationSendHoleAddress] == NUMBER_OF_TESTS_TO_RUN):
+        break
```
Aeliko: And this as well.

scripts/run_tests.sh
Aeliko, line 1: Could run this script in a [self-hosted runner](https://docs.github.com/en/actions/concepts/runners/self-hosted-runners) as part of the CI process.

src/test_battle.c
```diff
@@ -6,344 +6,25 @@
...
-    .weather = 0,
-    .fieldCondition = 0,
-    .terrain = 0,
-};
-
+static BOOL g_TestBattleCurrentComplete = FALSE;
+static BOOL g_TestBattleHasMoreTests = FALSE;
+static int g_CurrentTestIndex = TEST_START_INDEX;
+// TOSO: there is definitely some better way to do this, so that we don't need to worry if somehow this address is used by something else
+int *g_EmulatorCommunicationSendHole = (int*)0x023df15c;
```
Aeliko: Also I don't think this is good C. Please fix this before merging.

BluRosie: is there a reason it's 23df15c?  assign its address in rom.ld (can even make it some area that's unused now that we've replaced the function in arm9 or something), then just `extern int *g_EmulatorCommunicationSendHole;` would link that properly

Aeliko: There is absolutely no reason. I just compiled, debug printed the address, then hardcoded it.

scripts/run_tests.py
BluRosie, line 1: will do a better way for g_EmulatorCommunicationSendHoleAddress in a bit

src/battle/battle_script_commands.c
```diff
@@ -4400,6 +4407,29 @@ BOOL btl_scr_cmd_10C_gotoifterastallized(void *bsys UNUSED, struct BattleStruct
...
+    }
+
+    if (ctx->battlemon[battlerId].type1 == TYPE_FLYING) {
+        ctx->battlemon[battlerId].type1 = ctx->battlemon[battlerId].type2;
+    } else {
+        ctx->battlemon[battlerId].type2 = ctx->battlemon[battlerId].type1;
+    }
+
+    return FALSE;
+}
```
BluRosie, line 4442: assumption that roost user has flying type is not a valid assumption

src/individual/ServerFieldConditionCheck.c
```diff
@@ -1233,6 +1233,25 @@ void ServerFieldConditionCheck(void *bw, struct BattleStruct *sp) {
...
+                    if (sp->oneTurnFlag[i].roostFlag) {
+                        int species = PokeOtherFormMonsNoGet(sp->battlemon[i].species, sp->battlemon[i].form_no);
+                        u32 type1 = PokePersonalParaGet(species, PERSONAL_TYPE_1);
+                        u32 type2 = PokePersonalParaGet(species, PERSONAL_TYPE_2);
+
+                        // Assume vanilla situations
+                        // Handle Soak
+                        if (sp->battlemon[i].type1 == TYPE_WATER && sp->battlemon[i].type2 == TYPE_WATER) {
+                            break;
+                        }
```
BluRosie: i'll allow it but don't particularly like it

BluRosie (approved):

src/battle/battle_script_commands.c
Aeliko: You're right. Need to add Flying type check at line 4419 as well.

src/individual/ServerFieldConditionCheck.c
Aeliko: We should probably add some new move flags for Soak, Trick-or-Treat, Forest's Curse.

BluRosie (approved):

PR #622, opened 2025-12-19
accelrock animation

keruds:
I didn't see an animation so I made this up.

https://github.com/user-attachments/assets/15e70499-5cd1-4da6-9243-3cb4d596e6b4


BluRosie:
looks good to me!  script is a little wonky with its particle loads (loadparticlefromspa covers a lot of the loadsprite commands), but it working makes it less concerning to me :grin:

PR #630, opened 2025-12-23
add bytereplacement to allow for shop prices with 5 digits

zebben:
before:
image

after:
image


BluRosie:
note for the future:  https://raw.githubusercontent.com/pret/pokeheartgold/3d8189a6ea33e97138d54aacb5f9b97d3a15d908/asm/overlay_31.s

(just in case we ever need to wonder where this is from)

BluRosie (approved):

PR #639, opened 2025-12-30
add print*Message battle script commands, tryUseHeldItem, belch

kumakuma1:
hook print*message
hook tryUseHeld item
add Belch + Flag + fling interaction
move some arrays, make them accessable as extern variable


kumakuma1:
> sorry, what do we need the print*message commands for?

We tried to log the battle messages from there, but did not work...

kumakuma1:
> now fix ripen and harvest 😈
> 
> will review soon

Ripen is not in. I just left a comment where the HP recovering part could be handled so that it could work with a pop-up
.

PR #654, opened 2026-01-12
Flying Press, Freeze Dry (+Ring Target & Tera cleanup)

Pixelstyx:
Credit to @kumakuma1 for the framework in https://github.com/BluRosie/hg-engine/pull/516.

I think there's a bit of leftover code from Blu but it seems to just make it easier to implement fakemon?


Pixelstyx:
image

PR #657, opened 2026-01-14
hook battlerCheckSubstitute, clean up some move flags, add terrain pulse, gem order/message

kumakuma1:
remove from rom.ld

https://github.com/user-attachments/assets/1132911b-f7cd-453d-817b-167ba9c84873

oddish is grass/rock


BluRosie:
going to merge then make sure that subscripts are all set

PR #659, opened 2026-01-14
implement simultaneous HP bar animations

zebben:
- Simultaneous HP bar animations for spread moves
- Batched effectiveness messages for spread moves

https://github.com/user-attachments/assets/c5c77332-6190-4c52-9c59-15155149ac3c


data/battle_scripts/subscripts/subscript_0021_MOVE_FOLLOWUP_MESSAGE.s
```diff
@@ -3,6 +3,7 @@
 .data
 
 _000:
+    CompareVarToValue OPCODE_FLAG_SET, BSCRIPT_VAR_MOVE_STATUS_FLAGS, MOVE_STATUS_NO_MORE_WORK, _104
```
zebben: no-op this if we explicitly set `MOVE_STATUS_NO_MORE_WORK` - could this be a problem anywhere else? can add a different value if so

src/battle/battle_controller_player.c
```diff
@@ -83,3 +85,28 @@ BOOL LONG_CALL BattleContext_Main(struct BattleSystem *bsys, struct BattleStruct
...
+        if (ctx->multiHitCountTemp != 0) {
+            if (ctx->fainting_client != BATTLER_NONE || ctx->multiHitCount == 1 || ctx->waza_status_flag & 0x4000) {
+                ret = TRUE;
+            }
+        } else {
+            ret = TRUE;
+        }
+    }
+
+    if (ret && ctx->server_status_flag & SERVER_STATUS_FLAG_SIMULTANEOUS_DAMAGE) {
```
zebben, line 103: only change from vanilla is stashing the status flag for simultaneous damage here.

data/battle_scripts/subscripts/subscript_0021_MOVE_FOLLOWUP_MESSAGE.s
zebben: Aero pointed out that it could be. I'll add a new flag to replace this soon

src/battle/battle_script_commands.c
```diff
@@ -4381,14 +4402,158 @@ BOOL btl_scr_cmd_10C_gotoifterastallized(void *bsys UNUSED, struct BattleStruct
...
+        return FALSE;
+    }
+
+    // 0 = player; 1 = wild; 2 = enemy
+    int msgTargetType = 0;
+    if (IsClientEnemy(bsys, list[0])) {
+        msgTargetType = (BattleTypeGet(bsys) & BATTLE_TYPE_TRAINER) ? 2 : 1;
+    }
+
+    int baseMsgId = (flag & MOVE_STATUS_FLAG_SUPER_EFFECTIVE) ? 1610 : 1634;
```
Aeliko: TODO: make them constants

PR #662, opened 2026-01-16
Totem Battles!

Pixelstyx:
Incomplete PR for Totem battles. Does the following:
- Allows the user to set a temp var in DSPRE to force WildBattleSp to launch a Totem battle instead.
- Adds a framework and template for the user to set custom data per-species for these encounters, including ability, held item, moves, pp, IVs and nature.
- Adds a framework and template for the user to set per-species stat boosts at the start of battle.
- Doubles the Totem Pokemon's weight at the start of battle, provided it is not already a Totem species.
- Prevents the player from escaping while in Totem battles.
- Prevents the player from capturing Totem Pokemon, including custom handling for messages.
- Removes the 'wild' prefix before the Totem's nickname.

TODO:
- Dynamically implement the 'Totem' prefix when buffering nicknames.
- Give the Totem a more sophisticated AI. Likely needs the new AI framework being worked on right now.
- Factor in SOS battles to said AI if/when those are implemented.


src/game_stats.c
BluRosie, line 1: move structure to header file (maybe a save one?) and delete this

tools/btx
BluRosie, line 1: delete file before merging in move-dev

src/encounter.c
```diff
@@ -0,0 +1,97 @@
...
+#include "../include/constants/game.h"
+#include "../include/constants/game_stats.h"
+#include "../include/constants/moves.h"
+#include "../include/constants/species.h"
+
+struct BattleSetup LONG_CALL *BattleSetup_New(u32 heapID, u32 battleFlags);
+void LONG_CALL BattleSetup_InitFromFieldSystem(BattleSetup *setup, FieldSystem *fieldSystem);
+void LONG_CALL ov02_02247F30(FieldSystem *fieldSystem, u16 mon, u8 level, BOOL shiny, BattleSetup *setup);
+int LONG_CALL BattleSetup_GetWildTransitionEffect(struct BattleSetup *setup);
+int LONG_CALL BattleSetup_GetWildBattleMusic(struct BattleSetup *setup);
```
BluRosie: prefer to use function declarations that are in headers elsewhere / add them to headers to do so

src/encounter.c
BluRosie, line 1: this file may be able to be a field overlay?  is it an overlay 2 function?

BluRosie (changes requested):

src/battle/battle_controller_player.c
```diff
@@ -35,6 +36,31 @@ void overrideItemUsage(struct BattleSystem *bsys, struct BattleStruct *ctx)
...
+
+    for (battlerId = 0; battlerId < bsys->maxBattlers; battlerId++)
+    {
+        if (ctx->playerActions[battlerId][0] == CONTROLLER_COMMAND_RUN_INPUT)
+        {
+            if (ctx->com_seq_no[battlerId] == SSI_STATE_11)
+            {
+                if (fight_type & BATTLE_TYPE_TOTEM)
+                {
+                    msg.id = 1726; // Can't flee this fight!
```
zebben: constant

data/battle_scripts/subscripts/subscript_0000_START_ENCOUNTER.s
```diff
@@ -2,67 +2,95 @@
...
     PrintGlobalMessage 967, TAG_NICKNAME_NICKNAME, BATTLER_CATEGORY_ENEMY_SLOT_1, BATTLER_CATEGORY_ENEMY_SLOT_2
+    GoTo _SendOutPokemonMessage
 
-_084:
+_TotemEncounter:
+    // You are challenged by Totem {0}!
+    PrintGlobalMessage 1723, TAG_NICKNAME, BATTLER_CATEGORY_ENEMY
+    Wait
+    WaitButtonABTime 30
+    DoubleWeightTotem BATTLER_ENEMY
```
zebben: maybe we could have a more generic Totem command which handles the stat boosts and stuff in addition to weight? think we could avoid having to have both subscript logic and C logic then to handle stat boosts per species

Pixelstyx: Honestly this is a much smarter idea and I'm not sure why it didn't occur to me. I'll get on this tonight.

src/encounter.c
Pixelstyx: SetUpAndStartWildBattle and related functions are in main.

BluRosie: i'll take a look

include/constants/species.h
```diff
@@ -1190,10 +1190,13 @@
 #define SPECIES_MIMIKYU_BUSTED_LARGE (SPECIES_ALOLAN_REGIONAL_START + 28)
 #define SPECIES_KOMMO_O_LARGE (SPECIES_ALOLAN_REGIONAL_START + 29)
 
+#define MAX_TOTEM_FORMS_CANONICAL ((SPECIES_KOMMO_O_LARGE - SPECIES_RATICATE_ALOLAN_LARGE) + 1)
+// Add the amount of custom Totem forms to this constant to make sure they use the appropriate weight in battle. 
+// Don't bother making custom forms unless you want them to be caught.
+#define MAX_TOTEM_FORMS (MAX_TOTEM_FORMS_CANONICAL)
```
BluRosie: what is the difference between this and `NELEMS(TotemSpecies)`?  one of the recent pushes is to make it so that all overworld tags will always remain the same so that folks can always use an overworld tag down the line.  but adding a new form in the middle of the list would shift all of the later form tags

include/trainer_data.h
```diff
@@ -9,4 +9,6 @@ typedef enum TrainerGender {
     TRAINER_DOUBLE,
 } TrainerGender;
 
+typedef struct BattleSetup BattleSetup;
```
BluRosie, line 19: what do you need this here for, sorry?

BluRosie:
would you like me to look into why the tests are failing?  i think the message buffering stuff is just always happening

include/trainer_data.h
Pixelstyx: oh i don't think this needs to exist anymore. This PR has been around for so long that someone else added BattleSetup to battle.h.

Pixelstyx: ok no trying to move it causes a bunch of issues

Pixelstyx: do you want me to try and move it somewhere? will probably be a pain, i don't remember wtf i did in the first place to put it here

include/constants/species.h
Pixelstyx: Totem Forms will always be Totem Species, but not all Totem Species need to be Totem Forms. Basically to sidestep forcing the user to make custom forms for every totem, they can declare a regular species in the TotemSpecies list and it will have its weight temporarily doubled instead of using the form weight.

You should add the forms at the end, not in the middle, I just wanted to keep that const with the rest of the defines. You would just start counting a second series of IDs there, like how Pokeballs + Apriballs work.

include/trainer_data.h
BluRosie: what's the issues?  typically could just include battle.h where it's needed

Pixelstyx: let me give it a go

Pixelstyx: I think it creates cyclical dependency or something? It causes a lot of things in battle.h to become unrecognized when i try to include it in task.h.

BluRosie (approved):

BluRosie:
will merge on test completion

PR #682, opened 2026-02-03
Add more tests for type changing moves

Aeliko:


src/individual/MoveHitDefenderAbilityCheck.c
```diff
@@ -66,6 +66,7 @@ BOOL MoveHitDefenderAbilityCheckInternal(void *bw, struct BattleStruct *sp, int
             && (sp->battlemon[sp->defence_client].condition2 & STATUS2_SUBSTITUTE) == 0
             && (sp->multiHitCount <= 1)) // don't activate until the last hit of a multi-hit move
         {
+            debug_printf("Color Change\n");
```
BluRosie, line 69: remove debug print

PR #689, opened 2026-02-10
Scale shot, fix some stat change animations

kumakuma1:
https://github.com/user-attachments/assets/92b841de-c1ea-4457-92be-61616f44a720

Activation/message is skipped if stats are at max/min.
Add this skipping for fell stinger


armips/include/config.s
BluRosie, line 1: smh

include/config.h
BluRosie, line 1: smh x2

PR #693, opened 2026-02-16
Fix for doubles missingno. Hopefully pave the way for totems, raids, etc.

zebben:
Trainer 1v2 fight
https://github.com/user-attachments/assets/cc1f8f82-33fe-4c7e-a826-43468d0244e5

Wild 1v2 fight
https://github.com/user-attachments/assets/4527c44d-5187-4d87-be78-c06f4f8ee3b8

Basically just intercepting a bunch of different sprite loading functions to skip loads. Easier fix was to skip emit calls but that resulted in task corruption causing a crash after the battle finishes. Tested this with regular 2v2 and that works as well. Right now "invalid" is all driven from `IsBattlerSlotValid`.

`BattleControllerPlayer_GetBattleMon` is there for two reasons. First, it removes the shiny flag that missingno always has so we avoid the flash from that. This just worked out to be the . Second, it sets up `no_reshuffle_client` so that missingno does not get to  act.

The idea going forward is to modify `IsBattlerSlotValid` to check on the battle type and make slots "invalid" according to that logic then this stuff will skip the graphics rendering.


Aeliko (commented):
TODO: send out message

BluRosie (approved):

PR #694, opened 2026-02-17
Check exact battle messages in tests

Aeliko:


armips/data/moves.s
```diff
@@ -7185,7 +7185,7 @@ movedata MOVE_CHATTER, "Chatter"
     type TYPE_FLYING
     accuracy 100
     pp 20
-    effectchance DEBUG_NEEDS_TESTING ? 100 : 0 // always confuse
+    effectchance 0 // always confuse
```
BluRosie: chatter should always confuse now

PR #713, opened 2026-03-17
fix mints et al. and simplify capsule/patch back to old hook behavior

zebben:


src/pokemon.c
```diff
@@ -1485,16 +1485,6 @@ u32 LONG_CALL CheckIfMonsAreEqual(struct PartyPokemon *pokemon1, struct PartyPok
 BOOL LONG_CALL CanUseItemOnMonInParty(struct Party *party, u16 itemID, s32 partyIdx, s32 moveIdx, u32 heapID) {
     struct PartyPokemon *mon = Party_GetMonByIndex(party, partyIdx);
 
-    if (itemID == ITEM_ABILITY_CAPSULE)
```
zebben, line 1488: these are covered by `UseItemMonAttrChangeCheck`

src/party_menu.c
```diff
@@ -306,6 +306,10 @@ int LONG_CALL PartyMenu_HandleUseItemOnMon(struct PartyMenu *partyMenu)
         return PARTY_MENU_STATE_SELECT_MOVE;
     }
 
+    if (UseItemMonAttrChangeCheck(partyMenu, itemData) == TRUE) {
```
zebben, line 309: should more closely match the previous asm hook

BluRosie (approved):

PR #732, opened 2026-04-03
Disable Critical Health Bleeping

blackisme1:
I cant reproduce the 4 bleeps then fading away behavior from Gen VI onwards so its not an exact modernization but goddamn is it way better on the ears

https://github.com/user-attachments/assets/89297ef6-d562-4eeb-bddb-d3a0e0287634

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -218,4 +218,8 @@
 // add or remove mythicals as you see fit
 #define VANILLA_MYTHICALS
 
+// DISABLE_CRITICAL_HP_WARNING should be used if you want to disable the warning whenever your pokemon is at critical health
+// comment out this line if you do not want this to happen
+//#define DISABLE_CRITICAL_HP_WARNING
+
 #endif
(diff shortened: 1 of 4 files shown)


blackisme1:
I cant add a "reverse" config because battleSystem->criticalHpMusic doesnt exist (its been blanked over in the struct)

include/config.h
```diff
@@ -218,4 +218,8 @@
 // add or remove mythicals as you see fit
 #define VANILLA_MYTHICALS
 
+// DISABLE_CRITICAL_HP_WARNING should be used if you want to disable the warning whenever your pokemon is at critical health
+// comment out this line if you do not want this to happen
+#define DISABLE_CRITICAL_HP_WARNING
```
Aeliko, line 223: Default should be off

Aeliko (changes requested):

include/config.h
blackisme1: Shouldnt the default be modernization i.e. closest to "Non looping behavior"

Aeliko: OK fair enough

Aeliko (approved):

Issue #749, opened 2026-04-11
Red Card does not show its text on activation


kumakuma1:
Furret frisked Chikorita and found its Red Card!

Marshtomp used Bulldoze! ✅
It’s not very effective on the opposing Chikorita.
Furret’s Speed fell! ✅
**The opposing Chikorita’s Speed fell! ✅
The opposing Furret’s Speed fell! ✅
It’s super effective on the opposing Chikorita!
Treecko was dragged out!  ✅**
The opposing Furret is switched out by the Eject Pack! ✅
Rival  sent out Litten!
The opposing Litten’s Intimidate cuts Treecko’s Attack! ✅
The opposing Litten’s Intimidate cuts Furret’s Attack!
Furret used Sleep Talk!
But it failed!
The opposing Chikorita used Sleep Talk! ✅
[Pass] Eject Pack - Red Card does not prevent Eject Pack

kumakuma1:
#753

Issue #748, opened 2026-04-11
Phantom Force does not hit through Protect


kumakuma1:
Dreepy used Phantom Force!
Dreepy vanished instantly! ✅
The opposing Incineroar used Bite!
The attack of the opposing Incineroar missed! ✅
The opposing Incineroar used Protect!
The opposing Incineroar protected itself! ✅
move 182 damage roll 0
Dreepy used Phantom Force!
The opposing Incineroar protected itself!
expectation[3] ❌
[Known Failing] Phantom Force - Semi Invun, hit through Protect

PR #753, opened 2026-04-21
Fix Red Card text, moves that hit through Protect

kumakuma1:
#748 
#749


BluRosie (approved):

BluRosie:
will merge on build success

PR #755, opened 2026-04-21
Champions Sleep Modernization

blackisme1:
1, 2, 3 turns of sleep -> 1, 2, 2 turns of sleep


kumakuma1:
config please

blackisme1:
Idk if other modernizations even have configs, like sleep's gen 5 modernization to 1-3 turns from 1-4 turns doesnt have a config

kumakuma1:
Ideally you would add this into the config as well!

data/battle_scripts/subscripts/subscript_0018_FALL_ASLEEP.s
```diff
@@ -87,6 +87,9 @@ _182:
     PlayBattleAnimation BATTLER_CATEGORY_SIDE_EFFECT_MON, BATTLE_ANIMATION_ASLEEP
     Wait
     Random 2, 2
+    CompareVarToValue OPCODE_LTE, BSCRIPT_VAR_CALC_TEMP, 3, _lowerThanThreeTurns
+    UpdateVar OPCODE_SET, BSCRIPT_VAR_CALC_TEMP, 3
+_lowerThanThreeTurns:
```
Aeliko, line 92: Note:


In Pokémon Champions, sleep now lasts 2-3 turns (with an uneven distribution compared to previous generations, comparatively 2, 3, 4 turns vs. 2, 3, 3 turns), meaning it prevents movement for 1-2 turns.

So not just doing `Random 2, 1` is actually correct

Aeliko (approved):

Issue #770, opened 2026-05-01
Bad Egg faints in doubles with spread moves still observed


zebben:
https://discord.com/channels/446824489045721090/1201213967389966378/1499826111008342189

Issue #769, opened 2026-05-01
Pokeathlon expansion only exposes natdex shops


zebben:
each day has 2 shops. one for pre-natdex and one for post-natdex and we are only exposing post-natdex at the moment

PR #771, opened 2026-05-02
fix yet another bad egg

zebben:
pre-natdex pokeathlon shops

addresses #770 
addresses #769


BluRosie:
this one would be fine to merge if not for the linting changes.  preference is to roll it out soon in one big pr so that it doesn't become a burden on maintenance

zebben:
> this one would be fine to merge if not for the linting changes. preference is to roll it out soon in one big pr so that it doesn't become a burden on maintenance

reverted then pushed without lint

BluRosie:
ugh i misclicked and it did everything let me fix that

BluRosie (approved):

PR #779, opened 2026-05-06
Fix turning disable low health warning config off causing it to malfunction

blackisme1:
I guess those shouldnt be unused, because they are used when the config is disabled


BluRosie:
i'll just add a bytereplacement if the config is off that should undo it

PR #792, opened 2026-05-18
move imposter flag set to below the clear flag call

zebben:
tests pass locally

https://github.com/user-attachments/assets/9d362d5d-0666-4829-94ad-6c654bd561ea


BluRosie:
will see how build does and if it passes locally or not

PR #824, opened 2026-06-09
migrate to GCC for assembler (and a few other qol things)

BluRosie:


BluRosie:
just slightly under 3000 files!  no 3000+ here!

there should not be any crazy changes in the data/code files.  just mostly need a review for if it's okay that "`../../include/`" is removed from includes in src files, `#include` can be used in assembly files now (but not mixed, notably), and how things look

will be the pr that mandates the linter too

armips/asm/overworlds.s
```diff
@@ -41,38 +41,38 @@ nop
...
-
-@@_getDimorphism:
-    ldr r1, =0x020FECAE
-    ldrb r0, [r1, r0]
-    pop {r3, pc}
-
-.pool
-
-.endarea
+//.org 0x0206A338
```
zebben, line 44: this isn't needed anymore?

include/constants/item.h
```diff
@@ -2752,6 +2754,16 @@ enum ItemGeneration {
     GEN8,
     GEN9
 };
+#else
+#define CUSTOM 0
+#define GEN4 1
+#define GEN5 2
+#define GEN6 3
+#define GEN7 4
+#define GEN8 4
```
zebben: 5

include/constants/item.h
```diff
@@ -2752,6 +2754,16 @@ enum ItemGeneration {
...
+#define GEN9 5
```
zebben: 6

scripts/build_learnsets.py
```diff
@@ -19,14 +19,17 @@
...
+    baseSpeciesMaxIndex = 0
     with open(file_path) as f:
         for line in f:
             if len(line.split()) > 1:
                 test = line.split()[1].strip()
                 if 'SPECIES' in test and not '_START' in test and not '_SPECIES_H' in test and not '_NUM (' in line and not 'MAX_' in test:
                     species_dict[test] = index
                     index += 1
-    return species_dict
+                elif 'MAX_CANONICAL_MON_NUM' in test:
```
zebben, line 30: does this script still handle fakemon properly (referencing canonical count not actual)?

armips/asm/overworlds.s
BluRosie: it's usurped by https://github.com/BluRosie/hg-engine/blob/move-dev/data/SpeciesToOWFormFemale.c

scripts/build_learnsets.py
BluRosie: this is what NUM_MONS was in armips

zebben (approved):

BluRosie:
you would have merge conflicts (derogatory)

BluRosie:
inshallah tests pass, then constants can be fully consolidated and this can be merged

PR #836, opened 2026-06-20
allow all dumps to complete if one fails

zebben:
- wrap each dump in a try/catch
- hidden items further standardized
- fix for dumping species when not all species were present in base (i.e. new megas and such were added)
- fix for dumping trainers with randomized order
- fix headbutt dump

failure looks like:
```
Failed to dump Moves.c: 'Unknown key: a/9/9/9'
Completed with 1 failed dump(s): Moves.c
make[1]: *** [dump_narc_data] Error 1
make: *** [dumprom] Error 2
```

but still outputs the other files

### DIFF
--- a/data/Headbutt.c
+++ b/data/Headbutt.c
@@ -5174,7 +5174,7 @@ const HeadbuttArchiveData __data =
 
     .route34 = {
         .normalTreeCount = 15,
-        .specialTreeCount = 1,
+        .specialTreeCount = 0,
         .normalSlots =
         {
             { SPECIES_HOOTHOOT, 9, 10 },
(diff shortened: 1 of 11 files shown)


data/Headbutt.c
```diff
@@ -5174,7 +5174,7 @@ const HeadbuttArchiveData __data =
 
     .route34 = {
         .normalTreeCount = 15,
-        .specialTreeCount = 1,
+        .specialTreeCount = 0,
```
zebben, line 5177: looking at pret I think I just missed the cmp here and it was off? because it seems all empty

BluRosie (approved):

PR #856, opened 2026-07-07
support nickname char array via consts

zebben:


BluRosie (commented):
need to dump nicknames as well then?  or is this already supported in the dump script?  i do not remember

zebben:
> need to dump nicknames as well then? or is this already supported in the dump script? i do not remember

they dump as an int array (not using consts). we could also swap to having trainerdatagen read actual strings as Aero suggested. that'll require users to do some conversion step though

BluRosie:
> they dump as an int array (not using consts). we could also swap to having trainerdatagen read actual strings as Aero suggested. that'll require users to do some conversion step though

support both 😈😈😈

zebben:
> > they dump as an int array (not using consts). we could also swap to having trainerdatagen read actual strings as Aero suggested. that'll require users to do some conversion step though
> 
> support both 😈😈😈

will take a look tonight🫡

BluRosie (approved):

PR #857, opened 2026-07-12
fix: change the form to the default form

Phosphorus-M:
I don't know, I was having some troubles with this
I understand that the form 3 doesn't exists, we have just mapped 0 forms and the defualt form (the form 0).

0 Default = Lycanroc Midday Form
1 Default = Lycanroc Midnight Form
2 Default = Lycanroc Dusk Form

I can be wrong, if it's the case just close this PR :sob:

### DIFF
--- a/data/Evolutions.c
+++ b/data/Evolutions.c
@@ -11132,7 +11132,7 @@ const EvolutionTable __data[MAX_SPECIES_INCLUDING_FORMS + 1] =
 
     [SPECIES_ROCKRUFF] = {
         .entries = {
-            { EVO_LEVEL_DAY, 25, MON_WITH_FORM(SPECIES_LYCANROC, 3) },
+            { EVO_LEVEL_DAY, 25, SPECIES_LYCANROC },
             { EVO_LEVEL_NIGHT, 25, MON_WITH_FORM(SPECIES_LYCANROC, 1) },
             { EVO_NONE, 0, SPECIES_NONE },
             { EVO_NONE, 0, SPECIES_NONE },


BluRosie:
what was the issue you were having?  i am struggling to see how it happens, because own tempo rockruff has its own entry in the evolution table

Phosphorus-M:
I'm going to try to explain this as clearly as possible because my English is terrible, mabye I can be redundant.

Basically, when Rockruff evolves, I get a crash because `MON_WITH_FORM(SPECIES_LYCANROC, 3)` doesn't exist, there is no form 3

The species exists, and it has two alternate forms. If you count the base species, there are three variants in total, but the form array should be indexed from 0 to 2, not 3. According to Bulbapedia, a normal Rockruff should evolve into form 0 during the day, which is the base species rather than an alternate form

I know that Own Tempo Rockruff has its own species entry, but this change only applies to the normal Rockruff, a normal Rockruff should never evolve into Dusk Lycanroc; it should evolve into Midday Form

Form 0 = Species = Lycanroc Midday Form
Form 1 = Lycanroc Midnight Form
Form 2 = Lycanroc Dusk Form

Normal Rockruff should evolve into:
* Lycanroc Midday Form
* Lycanroc Midnight Form

Own Tempo Rockruff should evolve into:

* Lycanroc Dusk Form

---

Current behavior:

Normal Rockruff evolves into:
* Crash (invalid form 3)
* Lycanroc Midnight Form

Own Tempo Rockruff evolves into:
* Lycanroc Dusk Form

This patch fixes that evolution issue for Rockruff.

Sorry I'm not too clear :pensive:

BluRosie (approved):

PR #858, opened 2026-07-12
Add tools for dumping eventdata, scr_seq, and text, then dump

qtoloza:
Broken up into two parts so y'all can easily decide if you don't want to include the actual dumped files.


BluRosie:
ooh, love this!

small nits:
- get eventdata out of `armips/data` as it shouldn't be touched by armips
- refactor scr_seq to dump to gcc-assemblable format instead of armips...  will be slightly more involved in the makefile but much better overall (and will significantly help getting rid of armips)

qtoloza:
Will try to get around to it later this week!

qtoloza:
See what you think now! This produces a byte-identical ROM compared to https://github.com/BluRosie/hg-engine/pull/861.

data/battle_scripts/subscripts/subscript_0349_HANDLE_TERRAIN_SEEDS.s
```diff
@@ -22,7 +22,7 @@ _raiseDefense:
     GoTo _end
 
 _raiseSpDefense:
-    UpdateVar OPCODE_SET, BSCRIPT_VAR_MESSAGE, STAT_SPECIAL_DEFENSE
+    UpdateVar OPCODE_SET, BSCRIPT_VAR_MESSAGE, STAT_SPDEF
```
BluRosie, line 25: preference to keep `STAT_SPECIAL_DEFENSE` probably

BluRosie:
dumped scripts should go into `data` instead of `asm` (as they are data files and not code, similar to move scripts)

i do not mean to disparage you at all--could we make the dumps into an opt-in?  like only if the directory exists or the base file or something?  i do not want to further force folks who like dspre to conform to hg-engine, would much rather texts/scripts/event data be something that users can continue to use dspre only.  i can look into this if you would like to be done, you've done a lot so far 😁

qtoloza:
Oops, my bad, should have guessed that that was the organizational intent.

Yeah of course, not disparaging at all! Where would you like the opt-in switch to operate? I'm happy to take a look at it today if I get a chance.

1. Remove the additional dumped texts/scripts/event data and leave it up to users to use the dump script?
2. A flag that gates packing the dumped data? And it would make an exception for e.g. `scr_seq_003_commonscript`, the pre-existing text files, etc.?

I put 2 forward because I think the major benefit of this whole PR is diff tracking on all the hard work people put into scripting. Having the dumped data maintained on repo encourages that and reduces conflicts when e.g. hg-engine tweaks item naming in text files. But it's really up to you and your vision of people's workflows - it's easy for me to implement my preference in my fork :).

data/battle_scripts/subscripts/subscript_0349_HANDLE_TERRAIN_SEEDS.s
qtoloza: I swapped because this was the only use of `STAT_SPECIAL_DEFENSE` over `STAT_SPDEF` that I could find. `STAT_SPECIAL_ATTACK` is also unused. Are you migrating to preferring the longer form references everywhere?

BluRosie:
1. yes
2. yeah essentially...  or the presence of a specific file/folder would trigger a different makefile stuff

BluRosie:
if i were completely selfish i would gladly tell people to completely abandon dspre outside of a coordinates tracker that you then copy into hg-engine.  i would recommend this workflow lol, but some people really like the gui things

qtoloza:
Yeah, I figured people could have it either way with the dump script - edit in dspre, dump to reveal the diff from vanilla, and compile within engine natively. But you have an infinitely better idea of how people actually approach working with hg-e!

I think I should defer to you about how to implement from here, most of the work is done and the remaining items require more insight into your preference/vision. Happy with whatever you choose :).

BluRosie:
alright, are you able to remove the dumped files and build them based on a flag in the makefile or similar?  i can go through and figure out a good way from there

qtoloza:
Had some time today and took a stab at it based on your comments. Building dumped scripts is opt-in and granular to the different types of dumps. I left the dumped scripts in the PR because I really think having them as an automatic diff reference is very helpful, but it's easy to delete the dumps if you really disagree.

BluRosie:
okay sick!  i will merge this in to item-dev and do some experimentation 🫡

BluRosie (approved):

PR #869, opened 2026-07-23
Wiki Sync

zebben:
Sets up documentation/wiki in the actual hg-engine repo as a source to sync to hg-engine's wiki pages. It will run on merge to main which touches either the wiki workflow file or documentation/wiki contents.

Benefit would be allowing others to contribute wiki material without needing contributor access. I ran a couple of tests on my fork and it seems to work


BluRosie (approved):

BluRosie:
send it

PR #877, opened 2026-07-28
overlay 129 savings

zebben:
* gate some new tables behind config options
* move 2 medium cold function calls to src/individual 
* expand sItemFieldUseFuncs instead of full replace

I think we could move roamer code to field but I didn't test it so I left it out for the moment

~1 KB overlay 129 saved with configs on
almost 3 with configs off

the overworld change adds safety around forms which don't have their own ow sprites.

see the video for an example. "Chikorita" is poltchageist antique and "Cyndaquil" is form 0. both work with the change

https://github.com/user-attachments/assets/70158b39-1265-49ab-b360-cb43881b8797


bytereplacement
BluRosie, line 1: if you are going to have these be behind configs, can we have reversion paths in bytereplacement?  that way if someone doesn't want to have the config anymore (for whatever reason), they can unapply their changes

BluRosie (changes requested):

bytereplacement
zebben: yeah I can add a bytereplacement to restore the original bytes in an else

zebben:
zzz

BluRosie:
merge on build success

BluRosie (approved):

PR #891, opened 2026-08-05
clang format the repository

BluRosie:
some slight changes to clang format file and precommit hook now just formats the entire repository


zebben:
Yessssss. Will review when I get home

.githooks/pre-commit
```diff
@@ -1,19 +1,19 @@
...
-echo "Running pre-commit checks..."
+echo "Running pre-commit formatter..."
 
 CLANG_FORMAT_BIN=clang-format
 if command -v clang-format-18 >/dev/null 2>&1; then
   CLANG_FORMAT_BIN=clang-format-18
 fi
 
-FILES=$(git diff --cached --name-only --diff-filter=ACM | grep -E '^(src/.+\.c|include/.+\.h)$' || true)
+FILES=$(find . -maxdepth 3 -type f | grep -E '^(./src/.+\.c|./include/.+\.h)$' || true)
```
zebben, line 14: seems fine. we may need to be better about keeping our clang-format versions in sync but I don't really have a problem with it

BluRosie: we definitely need to be better about it because it changes version to version

BluRosie:
now it won't even run on the pr :( https://github.com/BluRosie/hg-engine/actions/runs/31367582505/job/93389225294?pr=891

BluRosie:
now just run the linter over the repo as part of the workflow with the pre-commit hook file directly
an example of a failed one where src/battle/ability.c isn't entirely conformant: https://github.com/BluRosie/hg-engine/actions/runs/31397429595/job/93483678681?pr=891

Aeliko (approved):

zebben (approved):
LGTM. looks like an imposter test is failing tho

PR #903, opened 2026-08-20
Add ds-rom build option

Mixone-FinallyHere:
…_order

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.mch
 build/*
 base/*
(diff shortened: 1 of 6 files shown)


Aeliko:
Is there a reason why it can't be the default (better DSPRE compatibility)?

How about the yaml file for file ordering (in DSPRE), is that needed here as well?

BluRosie:
i'll make it the default if rust is installed w ndstool as a fallback

PR #920, opened 2026-08-28
Ow table

agathaGengar:
I reformatted it to not be wonky and added some comments so clang will ignore it


BluRosie:
merge on build success

BluRosie (approved):

PR #921, opened 2026-08-31
Magicroom

kumakuma1:


Aeliko (commented):
Need to update `StruggleCheck` as well?

kumakuma1:
> Need to update `StruggleCheck` as well?

StruggleCheck should correctly **not** fetch the choice item if magic room is up

Aeliko (approved):

BluRosie:
merge on check successes

PR #923, opened 2026-09-01
1vs2 mega with correct limits

kumakuma1:
old one was never merged


BluRosie:
ah my bad for not resolving it properly

kumakuma1:
> ah my bad for not resolving it properly

Actually was good, because it was botched :/

BluRosie (approved):

PR #926, opened 2026-09-01
fix a bug with loading a save with an alt form follower

zebben:
Fixes  `ARM9: data abort (0225F024)` on loading a save with an alt form follower by hiding the follower sprite. The sprite will be rebuilt on taking a step following existing working logic. This is definitely not ideal but I think the impact is pretty minimal

### Steps to Reproduce
1. Load the attached save file
2. Open bag and use the Rotom Catalog on Rotom to change it to any (non-bulb) form
3. Save
4. Reload the save
5. Crash

[test.zip](https://github.com/user-attachments/files/31532348/test.zip)


BluRosie:
just to make sure--does this crash in blackthorn gym?

also this may actually fix desmume overworld weirdness if what i think the culprit might be is.  this is consistently seen in the test environment if you let it run for a little bit

zebben:
> just to make sure--does this crash in blackthorn gym?
> 
> also this may actually fix desmume overworld weirdness if what i think the culprit might be is. this is consistently seen in the test environment if you let it run for a little bit

this branch works in Blackthorn Gym w/ Rotom Fan follower. move-dev crashes with ARM9: data abort (0225F024)

zebben:
seems to be consistent though. from what I could find, the issue is just loading a save with an alt form follower

BluRosie:
this will at least overall be fixed by getting rid of otherpoke

idk we will have to see if things go wrong afterwards

zebben:
> this will at least overall be fixed by getting rid of otherpoke
> 
> idk we will have to see if things go wrong afterwards

it's all alt forms as far as I can tell. even like Raichu Alolan etc. I first noticed this on the otherpoke branch but I think I got the overworlds all updated there

PR #935, opened 2026-09-13
Steel Beam

kumakuma1:


data/battle_tests/moves/steel_beam/magic_guard.c
Aeliko, line 1: This is contrasting Mind Blown right? Can we have a test for that?

data/battle_tests/moves/steel_beam/magic_guard.c
```diff
@@ -0,0 +1,102 @@
+// Test: Steel Beam - Recoil on Protect
```
Aeliko:
```suggestion
// Test: Steel Beam - Magic Guard does not negate recoil
```

data/battle_tests/moves/steel_beam/normal_hit.c
```diff
@@ -0,0 +1,103 @@
+// Test: Steel Beam - Recoil on Protect
```
Aeliko:
```suggestion
// Test: Steel Beam - basic implementation
```

Aeliko (changes requested):

data/battle_tests/moves/steel_beam/magic_guard.c
kumakuma1: Magic Guard prevents Recoil from Mind Blown/Steel Beam.

Aeliko (approved):

PR #945, opened 2026-09-17
Fix Charging Moves, Reduction Berry Activation

kumakuma1:


data/battle_tests/moves/fly/mid_fight_neutral_damage.c
```diff
@@ -0,0 +1,102 @@
...
+            { ACTION_NONE, 0 },
+            { ACTION_NONE, 0 },
+            { ACTION_NONE, 0 },
+            { ACTION_NONE, 0 },
+            { ACTION_NONE, 0 },
+        } },
+    .expectations = {
+        { .expectationType = EXPECTATION_TYPE_MESSAGE, .expectationValue.message = "The opposing Mew used Fly!" },
+        { .expectationType = EXPECTATION_TYPE_MESSAGE, .expectationValue.message = "The opposing Mew flew up high!" },
+        { .expectationType = EXPECTATION_TYPE_MESSAGE, .expectationValue.message = "The opposing Mew became fully charged due to its Power Herb!" },
```
Aeliko, line 99: Should we check the attack message isn't shown again?

data/battle_tests/moves/seismic_toss/effectiveness_berry.c
```diff
@@ -0,0 +1,87 @@
...
+            { ACTION_NONE, 0 },
+            { ACTION_NONE, 0 },
+            { ACTION_NONE, 0 },
+            { ACTION_NONE, 0 },
+            { ACTION_NONE, 0 },
+            { ACTION_NONE, 0 },
+        } },
+    .expectations = {
+        { .expectationType = EXPECTATION_TYPE_MESSAGE, .expectationValue.message = "Primeape used Seismic Toss!" },
+        { .expectationType = EXPECTATION_TYPE_MESSAGE_DOES_NOT_CONTAIN, .expectationValue.message = "It's super effective!" },
```
Aeliko, line 87: We need to check that the berry doesn't activate and damage is not reduced in another test case, since we cannot have 2 negative test expectations

Aeliko (changes requested):

data/battle_tests/moves/fly/mid_fight_neutral_damage.c
kumakuma1: added another message with flame orb proc to check it

data/battle_tests/moves/seismic_toss/effectiveness_berry.c
kumakuma1: added a second iteration with different expectation

src/individual/BattleController_BeforeMove.c
```diff
@@ -2831,10 +2832,12 @@ BOOL BattleController_CheckTypeImmunity(struct BattleSystem *bsys, struct Battle
     }
     u32 flag = 0;
     int effectiveness = TYPE_MUL_NORMAL;
+    int target = ctx->moveTbl[ctx->current_move_index].target;
+    int power = ctx->moveTbl[ctx->current_move_index].power;
```
Aeliko, line 2836: Can we check for move split instead of power?

Aeliko (approved):

src/individual/BattleController_BeforeMove.c
kumakuma1: We ususally do check power for these things

